PR #1, opened 2019-09-17
Add style to make reveal animation more obvious

majido:
Changes:
 - Add style to make reveal animation more obvious
 - Move styles to a common file
 - Add 50vh padding at bottom to ensure the "pan" effect runs a bit longer


majido:
Actually I think I should remove that "50vh padding at bottom" since it shows the difference between clamped and unclamped cases

PR #2, opened 2020-02-24
Remove ScrollTimeline.fill attribute.

ogerchikov:
Updated based on spec change: [Removed fill from ScrollTimeline](https://github.com/w3c/csswg-drafts/pull/4750):
- Removed fill attribute.
- Return current time of 0 for before start phase and time_range for after end phase.

### DIFF
--- a/scroll-timeline-base.js
+++ b/scroll-timeline-base.js
@@ -123,7 +123,6 @@ export class ScrollTimeline {
       startScrollOffset: 'auto',
       endScrollOffset: 'auto',
       timeRange: 'auto',
-      fill: 'none',
 
       // Internal members
       animations: [],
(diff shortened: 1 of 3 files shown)


scroll-timeline-base.js
```diff
@@ -226,17 +224,11 @@ export class ScrollTimeline {
 
     // Step 3
     if (currentScrollOffset < startOffset) {
-      if (this.fill == 'none' || this.fill == 'forwards')
-        return unresolved;
       return 0;
```
flackr: This is not quite consistent with the resolution we arrived at - because 0 will leave the animation in the active phase whereas it should be just before the active phase. Perhaps for the polyfill we should return -epsilon? Note that we can actually return a different currentTime than the one we assign to the internal animation currentTime.

flackr:
I think this is obsolete now that c2b49d9de130008196e9578e443173d6f5abaf3d is landed?

Issue #3, opened 2020-03-25
Add unit/integration tests


majido:
Currently the polyfill does not have any tests.

I suggest we look into adding test and setting up a CI pipeline to check new commits against those tests.

## Tests

Should be fairly easy to add some basic tests around timeline time calculations and animations correctly running.

One idea is to import/adapt the [wpt tests](https://cs.chromium.org/chromium/src/third_party/blink/web_tests/external/wpt/scroll-animations) for ScrollTimeline and run them against the polyfill. I haven't fully thought this through but it seems possible albeit may require us to change tests in a way that we can't easily keep syncing them.

## Test harness and CI
Key requirement here is to be able to test against all mainstream browsers.

In the [past](https://github.com/majido/high-resolution-timestamp-polyfill/blob/master/karma.conf.js) I have had success using **Karma** testing framework in conjunction with **Sauce Labs** to run tests against multiple browsers. But this was some times ago and there may be better solutions now.

majido:
About the idea of reusing WPT idea. @flackr  pointed out that [web-animations-js polyfill](https://github.com/web-animations/web-animations-js/) is doing something like this.

The tricky bit is that we need to load the polyfill before the test runs. AFAICT, they are using Karma and [configure](https://github.com/web-animations/web-animations-js/blob/dev/test/karma-config.js#L18) that to load their [own special testharnessreport.js](https://github.com/web-animations/web-animations-js/blob/dev/test/resources/testharnessreport.js#L416
) that injects the polyfill using document.write() . This works because every wpt test loads testharnessreport.js at the top of the test.

Note that to use the above technique, there may be a need to change the polyfill so that it can be loaded synchronously though. From @flackr :

> One other point of contention is that the scroll-timeline polyfill right now loads as a module (read: asynchronously). I'm not attached to this - it seems as though it would be more useful to load synchronously into the main context to make it easier to load prior to running tests.

flackr:
@zouhir Is this complete now?

zouhir:
Yes, having I recall we agreed in a past Slack conversation that local unit-tests are redundant given we run WPT. Do you still think so?

flackr:
Yes, I think so! Thanks again for all your work here.

We can close the dependent testing bugs too right?

zouhir:
Yep!

PR #9, opened 2020-04-03
[WIP] Infrastructure work, build and CI setup, unit tests #3 

zouhir:
**PR status: still WIP**
___
# Bundling Source Files to a Synchronous IIFE Module

### Proposed solution
Compile the es-modules JavaScript source files with [microbundle](https://github.com/developit/microbundle) and produce an iife (Immediately invoked function expression) that synchronously appends ScrollTimeline constructor to the Window object.

### Why microbundle as a bundler of choice
This bundler module is capable of creating a small, optimized JS bundle from source files using rollup under the hood with Zero-configurations needed. 
It is currently used by some other project and polyfills such:

- [GoogleChromeLabs/css-paint-polyfill](https://github.com/GoogleChromeLabs/css-paint-polyfill/blob/master/package.json#L8)
- [GoogleChromeLabs/kv-storage-polyfill](https://github.com/GoogleChromeLabs/kv-storage-polyfill/blob/master/package.json#L9)
- [GoogleChromeLabs/audioworklet-polyfill](https://github.com/GoogleChromeLabs/audioworklet-polyfill/blob/master/package.json#L17)

# E2E Testing
Cypress.io has a nice API for writing automated browser test but it does not support Safari and support for FireFox is in Beta, that makes Saucelabs the best option for automated browser testing at this moment.

### 1: The Web-animations-js polyfill strategy
That polyfill have a mix of: JavaScript unit tests and e2e integration tests, all run in CI via Saucelabs which we are planning to use for ScrollTimeline. 

Unit tests written in Mocha and running against real browser APIs (not [JSDOM](https://github.com/jsdom/jsdom)) via [Karma test runner](https://karma-runner.github.io/latest/index.html).

I believe the e2e integration test setup (for WPT, Blink test files) is counter-intuitive and hard to maintain, for those reasons:
- Karma excels at running JS tests against DOM or web pages as a context, and not running tests in HTML files.
- Web-animations-js polyfills has a long chain of configurations: a [framework-specific task runner modules ](https://github.com/web-animations/web-animations-js/blob/dev/Gruntfile.js#L20) that pre-processes [target files](https://github.com/web-animations/web-animations-js/blob/dev/target-config.js) and has a modified testharnessreporter (forked from WPT) as well as custom Karma test harness runners you [can see here](https://github.com/web-animations/web-animations-js/tree/dev/test). Being as close as possible to WPT files is beneficial and the more pre-processing and customization, the harder it will be for us to maintain.

### 2: Philip Walton's Blog Post on using Saucelabs Rest API [read](https://philipwalton.com/articles/learning-how-to-set-up-automated-cross-browser-javascript-unit-testing/)

The solution outline in Philip Walton's (Engineer @ Google) Blog is more straight-forward than Web-animations-js polyfill solution, he uses Mocha (the client side library) to run tests inside the HTML pages, and post those pages to sauce labs using their REST API, all needed there is a custom "failLogger" when the tests are finished, which is 1 JS function that acts as an adapter between Mocha results and Saucelabs. 

One note is that the Saucelabs rest API does not support localhost:xxxx addresses, the developer in that post used [ngrok](https://ngrok.com/) and mentioned using Sauce Connect VPN tunnel if it's possible

Frame 1 (1)

### 3: (✅currently using) Using Webdriver with Saucelabs [read] (https://wiki.saucelabs.com/display/DOCS/Node.js+Test+Setup+Example)

I found connect WPT from lcoalhost:xxxx to Saucelabs via a sauce connect tunnel and using selenium web driver to test against a matrix of browsers is the simplest & most effective way to run WPT harness tests, and even do more if needed (screenshots, ref, etc).

Some of the pros:

- not much pre-processing needed: we only need to inject the polyfill script below the `WPT/**/*.html` test `</title>` 
- all tests are ran via WPT test runner, no customization to the runner or reporter.

### 4: WPT test runner --sauce flag

Monday, April 20, Stephen McGruer from the Chromium project highlighted on the slack channel that we can use --sauce-* flags, [they are documented here](https://web-platform-tests.org/running-tests/command-line-arguments.html?highlight=sauce) I believe if this works after we inject the HTML, will even be an improvements to saucelabs' Web Driver implementation

____
This is a child task of: https://github.com/flackr/scroll-timeline/issues/3


index.html
```diff
@@ -1,41 +1,23 @@
...
-  {transform: 'translateY(0)'},
-  {transform: 'translateY(' + (window.innerHeight - 100) + 'px)'},
-], {
-  duration: 1000,
-  timeline: st,
-});
-</script>
+<ul>
+    <li>
+        <a href="/demo/basic">demo/basic</a>
```
majido, line 13: Good idea on moving this into its own demo page.

index.html
```diff
@@ -1,41 +1,23 @@
...
-}
-</style>
+<!doctype html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Local Dev Demos</title>
```
majido, line 8: Why just local? I thought these demos will also work as expected on gh-pages.

package.json
```diff
@@ -0,0 +1,57 @@
...
+    "env": {
+      "test": {
+        "plugins": [
+          "@babel/plugin-transform-modules-commonjs"
+        ]
+      }
+    }
+  },
+  "homepage": "https://github.com/flackr/scroll-timeline#readme",
+  "devDependencies": {
```
majido, line 48: We now have 2K transitive package dependencies. 😮 

Almost all of these are dev dependencies but this was a bit just surprising to me. I guess bundling a whole web-server/testing framework (jest) and compiler/bundler (babel) does this. So I am not sure if there is a way to avoid this.

zouhir: hi Majid, we are removing `jest`, `babel` and the existing unit tests, and we will be using `WPT` tests, the WAAPI polyfill is in fact way worse than that: https://github.com/web-animations/web-animations-js/blob/dev/package.json

The new solution I am going to propose to you for testing WPT with polyfills should help reducing those dependencies. 

I look forward to updating the PR very soon & for your review & feedback!

PR #14, opened 2020-08-07
fixes #13

### LINKED ISSUE #13: Orientation isn't taken into account
```js
new ScrollTimeline({
    orientation: 'inline',
})
```

I'm not getting timeline ticks for horizontal scrolling areas

argyleink:
inline-scroll


src/scroll-timeline-base.js
```diff
@@ -288,7 +288,9 @@ export class ScrollTimeline {
     let timeRange = calculateTimeRange(this);
 
     // Step 2
-    let currentScrollOffset = this.scrollSource.scrollTop;
+    let currentScrollOffset = this.orientation === 'block'
```
flackr, line 291: Can you check for 'block' or 'vertical' and add a TODO similar to calculateScrollOffset (called above)?

argyleink: nice sure!

I've updated the PR to make inline scrolling the exception as well, felt more likely to be an exception than the rule. I tested and everything works great still.

Issue #16, opened 2020-10-23
Return currentTime of null when ScrollTimeline should be inactive.


flackr:
There are many tests currently failing with `Error: assert_equals: Sanity check the timeline is inactive`:
* Canceling an animation with inactive timeline should cause its start time and hold time to be unresolved 
* oncancel event is fired when the timeline is inactive. 
* Finishing idle animation attached to inactive timeline pauses the animation. 
* Finishing running animation attached to inactive timeline pauses the animation. 
* Pause pending task doesn't run when the timeline is inactive. 
* Animation start and current times are correct if scroll timeline is activated after animation.pause call. 
* Play pending task doesn't run when the timeline is inactive. 
* Animation start and current times are correct if scroll timeline is activated after animation.play call. 
* Animation start and current times are correct if scroll timeline is activated after setting start time. 
* Animation current time is correct when the timeline becomes newly inactive and then active again. 
* Setting the start time clears the hold time when the timeline is inactive 
* Setting an unresolved start time sets the hold time to unresolved when the timeline is inactive 
* Setting the start time resolves a pending ready promise when the timelineis inactive 
* Setting an unresolved start time on a play-pending animation makes it idle when the timeline is inactive 

Ensuring that we return a null currentTime for an inactive timeline should get us past these errors.

Issue #15, opened 2020-10-23
Support setting timeline on ProxyAnimation.


flackr:
Setting the timeline requires custom logic in a set timeline method in proxy-animation.js in order to correctly switch to/from ScrollTimelines and perform the appropriate changes to the animation expected by [setting-timeline.tentative.html](https://github.com/web-platform-tests/wpt/blob/master/scroll-animations/setting-timeline.tentative.html).

The following tests currently fail with `Error: promise_test: Unhandled rejection with value: object "TypeError: Failed to set the 'timeline' property on 'Animation': Failed to convert value to 'AnimationTimeline'."`:

* Setting a scroll timeline on a play-pending animation synchronizes currentTime of the animation with the scroll position. 
* Setting a scroll timeline on a pause-pending animation fixes the currentTime of the animation based on the scroll position once resumed                                                                                                                             
* Setting a scroll timeline on a reversed play-pending animation synchronizes the currentTime of the animation with the scroll position.                                                                                                                              
* Setting a scroll timeline on a running animation synchronizes the currentTime of the animation with the scroll position. 
* Setting a scroll timeline on a paused animation fixes the currentTime of the animation based on the scroll position when resumed 
* Setting a scroll timeline on a reversed paused animation fixes the currentTime of the animation based on the scroll position when resumed                                                                                                                           
* Transitioning from a scroll timeline to a document timeline on a running animation preserves currentTime 
* Transitioning from a scroll timeline to a document timeline on a pause-pending animation preserves currentTime 
* Transitioning from a scroll timeline to a null timeline on a running animation preserves the play state 
* Switching from one scroll timeline to another updates currentTime 
* Switching from a document timeline to a scroll timeline updates currentTime when unpaused via CSS. 
* Switching from a document timeline to a scroll timeline and updating currentTime preserves the new value when unpaused. 
* Switching from a document timeline to a scroll timeline and updating startTime preserves the new value when play is called. 

And one test fails with `Error: promise_test: Unhandled rejection with value: object "TypeError: Cannot set property timeline of #<t> which has only a getter"`:
* Switching from a null timeline to a scroll timeline on an animation with a resolved start time preserved the play state

bramus:
Fixed in #18 I believe.

PR #18, opened 2020-11-10
#15, #16: Implement support for setting a scroll timeline

kevers-google:
The primary goal of this PR was to add support for mutable timelines.  In the process of getting the mutable timeline tests to pass, several other issues were addressed.  Start and hold times were added to the proxy details in order to be able to manipulate them without affecting the play state of the underlying native animation.  A proxy is now used even if starting with a native animation + document timeline.  This change was necessary to enable hooking in set timeline support.  The pending playback rate is now tracked in the proxy details since inaccessible via the web-animations API.  Several API calls that were being routed to native implementation are now explicitly handled in the polyfill to prevent unwanted side effects due to changes to the underlying native animations.  As previously mentioned, changes to the play state were a potential side effect. Others include timing of when changes take effect such as the playback rate and start / hold times. 

With this patch the number of passing tests increases to 284.  Some notable areas of improvement are:
13 of 14 pass in scroll-animations/setting-start-time.html 
11 of 14 pass in scroll-animations/setting-timeline.tentative.html
2 of 4 pas in scroll-animations/scroll-animation-inactive-timeline.html


kevers-google:
Ready for review

src/proxy-animation.js
```diff
@@ -30,151 +30,470 @@ function createReadyPromise(details) {
...
+            details.holdTime = null;
+        }
         resolve();
       }
     });
   });
 }
 
+function getAnimationProperty(details, name) {
+  if (details.timeline && (name in details))
```
flackr, line 59: I'm a bit concerned about sometimes returning the property from details and sometimes from the animation. I think we should remove the name in details check as it should be constant for a given property whether it is stored in the proxy details or in the base animation.

src/proxy-animation.js
```diff
@@ -30,151 +30,470 @@ function createReadyPromise(details) {
...
+    if (playbackRate > 0 &&
+        currentTime >= details.animation.effect.getTiming().duration)
+      return true;
+
+    return false;
+  };
+
+  if (isFinished()) {
+    if (details.playState != 'finished') {
+      details.playState = 'finished';
```
flackr, line 106: Do we need to call removeAnimation? If not, and finished animations continue ticking, we shouldn't need to call addAnimation on line 113, right?

src/proxy-animation.js
```diff
@@ -30,151 +30,470 @@ function createReadyPromise(details) {
...
+    if (playbackRate < 0 && currentTime <= 0)
+      return true;
+    if (playbackRate > 0 &&
+        currentTime >= details.animation.effect.getTiming().duration)
+      return true;
+
+    return false;
+  };
+
+  if (isFinished()) {
```
flackr, line 104: nit: This nested if condition is a bit confusing, can we early return if the state didn't change? e.g.

const newPlayState = isFinished() ? 'finished' : 'running';
if (newPlayState == playState)
  return;
if (newPlayState == 'finished') {
  ...
} else {
  ...
}
details.playState = newPlayState;

src/proxy-animation.js
```diff
@@ -30,151 +30,470 @@ function createReadyPromise(details) {
...
-        animation: effect,
-        timeline,
-        playState: "idle",
-        readyPromise: null,
-        sequence: 0, /* Used to track ready promises. */
-        aborted: new Set(), /* Aborted sequences. */
-      });
-      return;
-    }
-    const isScrollAnimation = timeline && timeline instanceof ScrollTimeline;
```
flackr, line 56: Thanks for cleaning this up!!!

src/proxy-animation.js
```diff
@@ -30,151 +30,470 @@ function createReadyPromise(details) {
...
+      pendingPlaybackRate: null,
+      proxy: this,
       sequence: 0, /* Used to track ready promises. */
       aborted: new Set(), /* Aborted sequences. */
     });
   }
 
+  // -----------------------------------------
+  // Helper method
+  // -----------------------------------------
```
flackr, line 162: We shouldn't be exposing new methods on the animation object. Can we move this to an internal function like updateFinishedState? You can export this function to allow it to be called from other modules.

src/proxy-animation.js
```diff
@@ -30,151 +30,470 @@ function createReadyPromise(details) {
...
+
+  // -----------------------------------------
+  // Web animation API
+  // -----------------------------------------
+
   get effect() {
     return proxyAnimations.get(this).animation.effect;
   }
   set effect(newEffect) {
-    return proxyAnimations.get(this).animation.effect = newEffect;
```
flackr, line 72: Isn't the return necessary in order to make things like `a.effect = b.effect = effect` work?

src/proxy-animation.js
```diff
@@ -30,151 +30,470 @@ function createReadyPromise(details) {
...
+    proxyAnimations.get(this).animation.effect = newEffect;
   }
 
   get timeline() {
-    const internalTimeline = proxyAnimations.get(this).timeline;
-    if (internalTimeline !== undefined)
-      return internalTimeline;
-    return proxyAnimations.get(this).animation.timeline;
+    const details = proxyAnimations.get(this);
+    return details.timeline || details.animation.timeline;
```
flackr, line 185: Comment that if we set a null timeline we revert to the underlying animation's timeline.

src/proxy-animation.js
```diff
@@ -30,151 +30,470 @@ function createReadyPromise(details) {
...
+          details.holdTime = null;
+          break;
+      }
+      addAnimation(details.timeline, this);
+      if (pending)
+        createReadyPromise(details);
+    } else {
+      // TODO: polyfill mutable timeline support. Cannot assume the native
+      // animation supports mutable timelines. Could keep a list of detached
+      // timelines and pump updates to current time via rAF.
```
flackr, line 241: I think polyfilling mutable timelines is out of scope for this library. I.e. instead of a TODO we can just note that if the browser doesn't support mutable timelines this will fail. That being said, we can at least skip this if newTimeline is equal to details.animation.timeline right? This will allow it to succeed to go back to the original animation timeline.

src/proxy-animation.js
```diff
@@ -30,151 +30,470 @@ function createReadyPromise(details) {
...
+      details.animation.timeline = newTimeline;
+      if (fromScrollTimeline) {
+        // TODO: sync pending status & play state (ready promise).
+        details.timeline = null;
```
flackr, line 245: I think we should do this earlier, perhaps first thing in the else branch? If the assignment on 242 throws we might not run this right?

src/proxy-animation.js
```diff
@@ -30,151 +30,470 @@ function createReadyPromise(details) {
...
+    details.readyPromise = null;
+    if (typeof value == 'number') {
+      details.holdTime = null;
+      details.startTime = value;
+      details.playState = 'running';
+      const timelineTime = details.timeline.currentTime;
+      details.animation.currentTime =
+          (timelineTime - details.startTime) * this.playbackRate;
+      updateFinishedState(details);
+    } else {
```
flackr, line 274: What does not 'number' mean? undefined? null? can we explicitly test for this value with === rather than consulting the type and throw if we get the wrong type?

src/proxy-animation.js
```diff
@@ -30,151 +30,470 @@ function createReadyPromise(details) {
...
+      details.startTime = null;
+      details.playState =
+          (typeof previousCurrentTime == 'number') ? 'paused' : 'idle';
+    }
   }
 
   get currentTime() {
-    return proxyAnimations.get(this).animation.currentTime;
+    const details = proxyAnimations.get(this);
+    if (details.timeline) {
```
flackr, line 284: nit: Flip this and do an early return.

src/proxy-animation.js
```diff
@@ -30,151 +30,470 @@ function createReadyPromise(details) {
...
 
   get playbackRate() {
     return proxyAnimations.get(this).animation.playbackRate;
   }
   set playbackRate(value) {
-    proxyAnimations.get(this).animation.playbackRate = value;
+    const details = proxyAnimations.get(this);
+    details.animation.playbackRate = value;
+    details.pendingPlaybackRate = null;
+    if (details.timeline)
```
flackr, line 327: updateFinishedState is a noop if there is no timeline, we don't need this check.

src/proxy-animation.js
```diff
@@ -228,23 +544,32 @@ export class ProxyAnimation {
...
-    if (details.readyPromise)
-      return details.readyPromise;
+    if (details.timeline) {
+      if (details.readyPromise)
+        return details.readyPromise;
+
+      // TODO: If not waiting on a pending task, we still need to return a ready
+      // promise; however, the promise can be immediately resolved. Cannot use
+      // the underlying animation as it is intentionally locked in a
+      // pause-pending state.
```
flackr, line 554: This is just return Promise.resolve(); isn't it? Can we do this?

src/proxy-animation.js
```diff
@@ -228,23 +544,32 @@ export class ProxyAnimation {
...
 
 };
 
 export function animate(keyframes, options) {
   let timeline = options.timeline;
   if (!timeline || !(timeline instanceof ScrollTimeline)) {
-    return nativeElementAnimate.apply(this, [keyframes, options]);
+    let animation = nativeElementAnimate.apply(this, [keyframes, options]);
+    // Even through this animation runs as a native animation, we still wrap
+    // it in a proxy animation to allow changing of the animation's timeline.
```
flackr, line 567: Maybe this shouldn't be a special case anymore, i.e. this method can be

if (timeline instanceof ScrollTimeline)
  delete options.timeline;

// common animation construction.

if (timeline instanceof ScrollTimeline) {
  animation.pause();
  proxyAnimation.play();
}

src/scroll-timeline-base.js
```diff
@@ -160,15 +154,23 @@ export function removeAnimation(scrollTimeline, animation) {
...
- * Attaches a Web Animation instance to ScrollTimeline
+ * Attaches a Web Animation instance to ScrollTimeline.
  * @param scrollTimeline {ScrollTimeline}
  * @param animation {Animation}
  * @param options {Object}
  */
 export function addAnimation(scrollTimeline, animation, options) {
-  let animations = scrollTimelineOptions.get(scrollTimeline).animations;
-  animations.push(animation);
+  if (!find(scrollTimeline, animation)) {
```
flackr, line 170: I'm a bit worried this may cover up errors of cases where we should have removed the animation but failed to - could we throw an error if the animation already exists?

src/scroll-timeline-base.js
```diff
@@ -160,15 +154,23 @@ export function removeAnimation(scrollTimeline, animation) {
   animations.splice(index, 1);
 }
 
+function find(scrollTimeline, animation) {
+  const animations = scrollTimelineOptions.get(scrollTimeline).animations;
+  let index = animations.indexOf(animation);
+  return (index === -1) ? null : animations[index];
+}
```
flackr, line 161: I'm not sure it's worth having a dedicated function for a single use of find

src/scroll-timeline-base.js
```diff
@@ -160,15 +154,23 @@ export function removeAnimation(scrollTimeline, animation) {
   animations.splice(index, 1);
 }
 
+function find(scrollTimeline, animation) {
+  const animations = scrollTimelineOptions.get(scrollTimeline).animations;
+  let index = animations.indexOf(animation);
+  return (index === -1) ? null : animations[index];
```
flackr, line 160: The return value doesn't seem to be used - just the truthiness - should this be a contains method?

flackr (changes requested):
Overall this looks good. If there are spec references for some of the algorithms here can you add them in comments similar to the methods in scroll-timeline-base.js?

src/proxy-animation.js
kevers-google: start and hold time are examples of places where they will differ.  The underlying animation is locked in a pause pending state, so it will have a hold time but no start time.  Having said, that,  I can remove this method to make it clear where we are fetching the property from.  My reason for adding in the first place, was to cut down on repetitious code.  Having said that, possibly better to be clear.

src/proxy-animation.js
kevers-google: Can A import something from B and B import something from A?

kevers-google: Can use another design pattern to avoid adding a method to the animation object.

src/proxy-animation.js
kevers-google: AFAIK setters shouldn't return values.

src/scroll-timeline-base.js
```diff
@@ -31,16 +31,10 @@ function scrollEventSource(scrollSource) {
...
   if (animations.length === 0) return;
-  let currentTime = scrollTimelineInstance.currentTime;
+  let timelineTime = scrollTimelineInstance.currentTime;
+
   for (let i = 0; i < animations.length; i++) {
-    // The web-animations spec says to throw a TypeError if you try to seek to
-    // an unresolved time value from a resolved time value, so to polyfill the
-    // expected behavior we cancel the underlying animation.
-    if (currentTime == null) {
-      if (animations[i].playState === "paused") animations[i].cancel();
```
flackr, line 40: The purpose of canceling the animation here is to ensure that the animation effect stops taking effect when the animation timeline becomes inactive. I think we should continue to do this in order to get the correct visual state when the timeline goes inactive. Note that this may need to properly reflect the timeline state.

kevers-google:
Would like to defer the step by step comments to a later PR when there is closer alignment with the spec. Currently, we take liberties with the ready promise and play state.

flackr:
Sure, we can absolutely defer some of these improvements, I mostly wanted to ensure there are TODO's to track work that we know we need to fix later.

When running demo/parallax/ with your changes the animations seem to stop working after you scroll past their end point. Can we fix this so that the polyfill remains functional for demos using it?

kevers-google:
Demo fixed. The tickAnimation method was only ticking if playState was 'running' and should also tick if 'finished'.

kevers-google:
# of passing tests is up to 302 from 284. The reason is that the following test was failing due to multiple test timeouts and now runs:

scroll-animations/updating-the-finished-state.html

18 of the 29 tests in this file now pass.

src/proxy-animation.js
```diff
@@ -29,152 +29,518 @@ function createReadyPromise(details) {
...
+        details.animation.currentTime = previousCurrentTime;
+      }
+    });
+  }
+}
+
+function effectEnd(details) {
+  const timing = details.animation.effect.getTiming();
+  const totalDuration =
+     timing.delay + timing.endDelay + timing.iterations * timing.duration;
```
flackr, line 127: FYI activeDuration (used by calculateTargetEffectEnd) already accounts for iterations * duration.

src/proxy-animation.js
```diff
@@ -29,152 +29,518 @@ function createReadyPromise(details) {
...
+  return Math.max(0, totalDuration);
```
flackr, line 128: Is the max of 0 significant? Isn't this likely to result in some incorrect state for an animation whose effect ends before 0?

kevers-google:
Per the spec, the end time has a lower bound of 0.0:
  https://drafts.csswg.org/web-animations-1/#end-time

flackr (approved):
Thanks for the explanation, a spec link may be helpful to justify this but I'll go ahead and merge this since otherwise everything looks good.

PR #24, opened 2020-11-26
Support writing mode

kevers-google:
With this change, all tests now pass in:

scroll-animations/current-time-writing-modes.html


src/scroll-timeline-base.js
```diff
@@ -89,9 +115,14 @@ export function installScrollOffsetExtension(parseFunction, evaluateFunction) {
...
-  if (orientation === "block") orientation = "vertical";
-  else if (orientation === "inline") orientation = "horizontal";
+  // Only one horizontal writing mode: horizontal-tb.  All other writing modes
+  // flow vertically.
+  const horizontalWritingMode =
+    getComputedStyle(this.scrollSource).writingMode == 'horizontal-tb';
+  if (orientation === "block")
+    orientation = horizontalWritingMode ? "vertical" : "horizontal";
+  else if (orientation === "inline")
+    orientation = verticalWritingModel ? "vertical" : "horizontal";
```
flackr, line 125: verticalWritingModel is undefined. This probably points to a lack of tests covering this.

src/scroll-timeline-base.js
```diff
@@ -58,6 +58,32 @@ function calculateTimeRange(scrollTimeline) {
...
+ * @param scrollTimeline {ScrollTimeline}
+ * @param orientation {String}
+ * @returns {Number}
+ */
+function directionAwareScrollOffset(scrollSource, orientation) {
+  const style = getComputedStyle(scrollSource);
+  // All writing modes are vertical except for horizontal-tb.
+  // http://drafts.csswg.org/css-writing-modes-4/#block-flow
+  const horizontalWritingMode = style.writingMode == 'horizontal-tb';
+  let currentScrollOffset  = scrollSource.scrollTop;
```
flackr, line 73: Looks like sieways-lr switches the inline direction to be from bottom to top, which may require flipping or taking the absolute value of the scrollTop. See https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode#Examples. We can add a TODO for this - it also seems to be unsupported in chrome.

flackr (changes requested):

src/scroll-timeline-base.js
kevers-google: Added a TODO.

src/scroll-timeline-base.js
kevers-google: Oops.  Fixed.  Yes we have a hole in the test coverage.

flackr (approved):

PR #26, opened 2021-01-07
Fix playState getter

kevers-google:
Tiny patch to fix the scope of the details variable in the playState getter. Tests were able to pass even without marking details with const/let/var, since the assignment could default to the global scope.  In a strict environment this is not allowed.


flackr (commented):
Doesn't this require rebuilding the release?

kevers-google:
Ack.  I'll put together a new pull request.

flackr:
GitHub is saying there are merge conflicts, perhaps you need to sync your local branch to master and rebuild.

kevers-google:
Would like to abandon this pull request.  Have a new one with the conflicts resolved.

PR #32, opened 2021-06-02
Add CSSOM polyfill

kevers-google:

PR #33, opened 2021-06-03
Add CSSOM polyfill

kevers-google:
Polyfills the following:
CSSUnitValue
CSSKeywordValue
CSSMath*
CSS.<unit> methods

Replaces https://github.com/flackr/scroll-timeline/pull/32 which was not properly synced.


bramus:
Verified this one by adjusting my https://codepen.io/bramus/pen/PobZbBV?editors=1010 demo as follows:

- Remove import of `css-typed-om` + call of `polyfill()` _(lines 8-12)_
- Load the `scroll-timeline.js` from this PR using `https://rawcdn.githack.com/flackr/scroll-timeline/b24f30114a2026bb564d8177a5cee9f484a1c064/dist/scroll-timeline.js` as the `src` _(line 21)_

Nice! 😊

src/proxy-cssom.js
```diff
@@ -0,0 +1,211 @@
...
+    'Hz',
+    'kHz',
+    // Resolution
+    'dppx',
+    'dpi',
+    'dpcm',
+    // Other units
+    "fr"
+  ].forEach((name) => {
+    if (!CSS[name]) {
```
flackr, line 204: Rather than checking existence for, and optionally installing each type, should we just have all of this conditional on the existence of window.CSS or on the existence of CSSUnitValue? I suspect that if a browser didn't support 'dpcm' for example, but did support CSSUnitValue, that trying to construct a CSSUnitValue(value, 'dpcm') would just throw an error, and so exposing the type in the CSS object is potentially introducing errors through suggested support for units.

flackr (changes requested):

src/proxy-cssom.js
```diff
@@ -0,0 +1,211 @@
...
+      get unit() {
+        return  privateDetails.get(this).unit;
+      }
+
+      toString() {
+        const details = privateDetails.get(this);
+        return `${details.value}${displayUnit(details.unit)}`;
+      }
+    }
+    window.CSSUnitValue = ProxyCSSUnitValue;
```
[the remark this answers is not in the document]
kevers-google: Done

src/proxy-cssom.js
kevers-google: Done

src/proxy-cssom.js
```diff
@@ -0,0 +1,211 @@
...
+      switch(unit) {
+        case 'percent':
+          return '%';
+        case 'number':
+          return '';
+        default:
+          return unit.toLowerCase();
+      }
+    }
+    class ProxyCSSUnitValue {
```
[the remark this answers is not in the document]
kevers-google: Done

src/proxy-cssom.js
```diff
@@ -0,0 +1,200 @@
...
+      }
+    }
+  };
+
+  if (!window.CSS) {
+    if (!Reflect.defineProperty(window, 'CSS', { value: {} }))
+      throw Error(`Error installing CSSOM support`);
+  }
+
+  if (!window.CSS.CSSUnitValue) {
```
flackr, line 149: I think you meant window.CSSUnitValue?

flackr (changes requested):

src/proxy-cssom.js
kevers-google: done

flackr (approved):

PR #48, opened 2022-04-08
CSS Polyfill

mehdi-kazemi:
Parse style tags and create ScrollTimeline using the JS polyfill.


src/proxy-animation.js
```diff
@@ -568,9 +568,22 @@ function tickAnimation(timelineTime) {
...
     const timelineTimeMs = fromCssNumberish(details, timelineTime);
 
-    details.animation.currentTime =
-        (timelineTimeMs - fromCssNumberish(details, this.startTime)) *
-            this.playbackRate;
+    let currentTime =
+      (timelineTimeMs - fromCssNumberish(details, this.startTime)) *
+      this.playbackRate;
+
+    const EPS = 1e-3;
```
kevers-google, line 575: Defer to a separate PR.  The purpose of suggesting that you try this approach was to highlight the source of the problem for non-inclusive end-points.  I do not believe that this is the best way of accomplishing making the endpoints inclusive as it feels hacky.

src/scroll-timeline-base.js
```diff
@@ -492,11 +492,11 @@ export class ScrollTimeline {
 
     // Step 3
     if (currentScrollOffset < startOffset)
-      return CSS.percent(0);
+      return CSS.percent(-0.001);
```
kevers-google, line 495: Leave at 0 and 100% for this PR.  We can discuss how to handle before and after phase separately.  My preference is to eliminate the clamping entirely and rework scrollOfffsets.

src/scroll-timeline-css-parser.js
```diff
@@ -0,0 +1,327 @@
...
+
+    // If this sheet has no srcURL (like from a <style> tag), we are
+    // done. Otherwise, we have to find `url()` functions and resolve
+    // relative and path-absolute URLs to absolute URLs.
+    if (!srcUrl) {
+      return p.sheetSrc;
+    }
+
+    p.sheetSrc = p.sheetSrc.replace(
+      /url\(["']*([^)"']+)["']*\)/g,
```
kevers-google, line 46: Regex tend to be hard to read.  Suggest assigning to a const with a comment to better document what it is being used for.

src/scroll-timeline-css-parser.js
```diff
@@ -0,0 +1,327 @@
...
+    if (timelineName && !rule.selector.includes("@keyframes")) {
+      this.cssRulesWithTimelineName.push({
+        selector: rule.selector,
+        'animation-name': undefined,
+        'animation-timeline': timelineName
+      });
+    }
+  }
+
+  parseIdentifier(p) {
```
kevers-google, line 188: Did his parse code come from another project?  If so, can you cite the project.  Is the parse code bundled in such a manner than it can be reused without copying?

src/scroll-timeline-css-parser.js
```diff
@@ -0,0 +1,327 @@
...
+  lookAhead(s, p) {
+    return p.sheetSrc.substr(p.index, s.length) == s;
+  }
+
+  peek(p) {
+    return p.sheetSrc[p.index];
+  }
+}
+
+const identMatcher = /[\w\\\@_-]+/g;
```
kevers-google, line 303: const IDENT_MATCHER = ...

src/scroll-timeline-css.js
```diff
@@ -0,0 +1,192 @@
...
+    // TODO
+  }
+
+  document.querySelectorAll("style").forEach((tag) => handleStyleTag(tag));
+  document
+    .querySelectorAll("link")
+    .forEach((tag) => handleLinkedStylesheet(tag));
+}
+
+function getSourceElement(source) {
```
kevers-google, line 51: Worth adding a comment for which version of the spec this is based on as this is likely to change.  Add a TODO to update to new version of the spec.

src/scroll-timeline-css.js
```diff
@@ -0,0 +1,192 @@
...
+  if (matches) {
+    return document.getElementById(matches[1]);
+  } else if (source === "auto") {
+    return document.scrollingElement;
+  } else {
+    return null;
+  }
+}
+
+function convertOneScrollOffset(part) {
```
kevers-google, line 62: scrollOffsets are likely to be deprecated.  Similar to my previous comment, please add a comment to indicate version of the spec with TODO to update to latest spec as it is ratified.

src/scroll-timeline-css.js
```diff
@@ -0,0 +1,192 @@
...
+  }
+}
+
+export function initCSSPolyfill() {
+  initMutationObserver();
+
+  window.addEventListener('animationstart', (evt) => {
+    const anim =
+      evt.target.getAnimations().filter(
+        anim => anim.animationName == evt.animationName)[0];
```
kevers-google, line 170: Not safe to assume that there will be only one CSS animation on an element.

src/scroll-timeline-css.js
```diff
@@ -0,0 +1,192 @@
...
+    return scrollTimeline;
+  } else {
+    return null;
+  }
+}
+
+export function initCSSPolyfill() {
+  initMutationObserver();
+
+  window.addEventListener('animationstart', (evt) => {
```
kevers-google, line 167: Add a comment to indicate why we are capturing the animation this way.

src/scroll-timeline-css.js
```diff
@@ -0,0 +1,192 @@
...
+        anim => anim.animationName == evt.animationName)[0];
+
+    const timelineName = parser.getScrollTimelineName(anim.animationName, evt.target);
+
+    if (timelineName) {
+      const scrollTimeline = createScrollTimeline(timelineName);
+      // If there is a scrollTimeline name associated to this animation,
+      // cancel it, whether we create a new animation or not
+      // depends on the fact that whether that scrollTimeline was valid or not
+      anim.cancel();
```
kevers-google, line 179: This approach results in a programmatic animation and not a CSS animation.  As a result, composite-order will be incorrect. Does not need to be resolved in this PR, but please add a comment.

src/scroll-timeline-css.js
```diff
@@ -0,0 +1,192 @@
...
+
+function isDescendant(child, parent) {
+  while (child) {
+    if (child == parent) return true;
+    child = child.parentNode;
+  }
+  return false;
+}
+
+function getScrollOffsets(source, offsets) {
```
kevers-google, line 115: Add comment with TODO. scrollOffsets are likely to be deprecated.

src/scroll-timeline-css.js
```diff
@@ -0,0 +1,192 @@
...
+      const scrollTimeline = createScrollTimeline(timelineName);
+      // If there is a scrollTimeline name associated to this animation,
+      // cancel it, whether we create a new animation or not
+      // depends on the fact that whether that scrollTimeline was valid or not
+      anim.cancel();
+      if (scrollTimeline) {
+        if (anim.timeline != scrollTimeline) {
+          const target = anim.effect.target;
+          const keyframes = anim.effect.getKeyframes();
+          target.animate(keyframes, { timeline: scrollTimeline });
```
kevers-google, line 184: new ProxyAnimation(anim, scrollTimeline);

This way the native animation is still a CSS animation. Furthermore, we are not potentially creating a cancel event or a new animation start event.

src/scroll-timeline-css-parser.js
```diff
@@ -0,0 +1,360 @@
...
+        .exec(rule.block.contents)?.[1]
+        .trim();
+    } else if (hasAnimation) {
+      let shorthand = RegexMatcher.ANIMATION
+        .exec(rule.block.contents)?.[1]
+        .trim();
+
+      if (shorthand) {
+        let remainingTokens = removeKeywordsFromAnimationShorthand(shorthand);
+        // TODO we are assuming the first one that is remaining is
```
kevers-google, line 149: Should be easy enough to validate the names based on looking them up in @keyframes and @*-timeline rules.

src/scroll-timeline-css-parser.js
```diff
@@ -0,0 +1,360 @@
...
+  'q',
+  'in',
+  'pc',
+  'pt',
+  'px',
+  // Percentage.
+  '%',
+];
+
+export const RegexMatcher = {
```
kevers-google, line 328: Suggest moving constants near the top of the file.  Though we only need these constants defined before we use them, it is more difficult  to read the code if needing to frequently look ahead to see the definitions.

src/scroll-timeline-css.js
```diff
@@ -0,0 +1,184 @@
...
+    //   addProxyForElement(element);
+    // });
+  });
+
+  sheetObserver.observe(document.documentElement, {
+    childList: true,
+    subtree: true,
+  });
+
+  function handleStyleTag(el) { // el: HtmlStyleElement
```
kevers-google, line 32: Prefer using documentation format consistent with the closure compiler.
see https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler

[@param {Type} varname Description](https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#param-type-varname-description)

src/scroll-timeline-css-parser.js
```diff
@@ -0,0 +1,377 @@
...
```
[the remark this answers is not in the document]
mehdi-kazemi: It's imported in scroll-timeline-css.js

src/scroll-timeline-css-parser.js
```diff
@@ -0,0 +1,381 @@
...
```
kevers-google, line 23: Does this need to be exported?

src/scroll-timeline-css-parser.js
```diff
@@ -0,0 +1,381 @@
...
+    }
+
+    // If this sheet has no srcURL (like from a <style> tag), we are done.
+    // TODO: Otherwise, we have to find `url()` functions and resolve
+    // relative and path-absolute URLs to absolute URLs.
+    return p.sheetSrc;
+  }
+
+  getScrollTimelineName(animationName, target) {
+    for (let i = this.cssRulesWithTimelineName.length - 1; i >= 0; i--) {
```
kevers-google, line 104: Add comment why we are traversing backwards along with TODO for rule specificity.

src/scroll-timeline-css-parser.js
```diff
@@ -0,0 +1,381 @@
...
+    // TODO is it enough to check with "includes()"
+    const hasAnimationName = rule.block.contents.includes("animation-name:");
+    const hasAnimationTimeline = rule.block.contents.includes("animation-timeline:");
+    const hasAnimation = rule.block.contents.includes("animation:");
+
+    let timelineNames = [];
+    let animationNames = [];
+
+    if (hasAnimationTimeline) {
+      timelineNames = this.extract(rule.block.contents, RegexMatcher.ANIMATION_TIMELINE)
```
kevers-google, line 167: Code duplication.  Add a helper method. i.e.

extractMatches(regex) {
   return this.extract(...).split(...).map(...);
}

src/scroll-timeline-css-parser.js
```diff
@@ -0,0 +1,381 @@
...
+          selector: rule.selector,
+          'animation-name': animationNames[i % animationNames.length],
+          'animation-timeline': timelineNames[i % timelineNames.length]
+        });
+      }
+    }
+
+  }
+
+  extractAnimationName(shorthand) {
```
kevers-google, line 229: This and the following extract functions are nearly identical. They can call a common method that checks if a name is in a map. e.g.

extractAnimationName(shorthand) {
  return this, findMatchingEntryInMap(shorthand, this.keyframeNames);
}
findMatchingEntryInMap(shorthand, map) {
   const matches = shorthand.split(" ").filter(part => map.has(part));
   return matches ? matches[0] : null;
}

src/scroll-timeline-css.js
```diff
@@ -0,0 +1,194 @@
...
+        if (addedNode instanceof HTMLStyleElement) {
+          handleStyleTag(addedNode);
+        }
+        if (addedNode instanceof HTMLLinkElement) {
+          handleLinkedStylesheet(addedNode);
+        }
+      }
+    }
+
+    // TODO: Not a good idea to proxy all elements.
```
kevers-google, line 18: This comment implies that you currently ARE proxying all elements. 

We can simply proxy element.style similar to how we proxy element.animate.  We take only a minor hit if setting a non-animation style on an element.  We accomplish this by swapping out Element.prototype.style.

src/scroll-timeline-css.js
```diff
@@ -0,0 +1,194 @@
...
+        // cancel it, whether we create a new animation or not
+        // depends on the fact that whether that scrollTimeline was valid or not
+
+        // Current approach (cancelling current animation and creating a new one) results in a
+        // programmatic animation and not a CSS animation.
+        // As a result, composite-order will be incorrect.
+        // TODO: Don't cancel animation, pause and create a proxy for it instead,
+        // this way the native animation is still a CSS animation.
+        // Furthermore, we are not potentially creating a cancel event or a new animation start event.
+        anim.cancel();
```
kevers-google, line 180: No need to cancel the animation.  In fact, the cancel is being done without checking if there is a scrollTimeline.

src/scroll-timeline-css.js
```diff
@@ -0,0 +1,194 @@
...
+        if (scrollTimeline) {
+          if (anim.timeline != scrollTimeline) {
+            const target = anim.effect.target;
+            const keyframes = anim.effect.getKeyframes();
+            target.animate(keyframes, { timeline: scrollTimeline });
```
kevers-google, line 185: const proxy = new ProxyAnimation(anim);
proxy.timeline = scrollTimeline;

kevers-google: new ProxyAnimatoin(anim, scrollTimeline);

src/scroll-timeline-css-parser.js
```diff
@@ -0,0 +1,381 @@
...
+    }
+
+    if (hasAnimationTimeline && hasAnimationName) {
+      this.saveRelationInList(rule, timelineNames, animationNames);
+      return;
+    }
+
+    if (hasAnimation) {
+      this.extract(rule.block.contents, RegexMatcher.ANIMATION)
+        .split(",").map(shorthand => shorthand.trim()).forEach(shorthand => {
```
kevers-google, line 183: getParts(shorthand).forEach(...)

src/scroll-timeline-css-parser.js
kevers-google: Comment that also used in scroll-timeline-css.js

kevers-google (approved):
Looks good.

PR #54, opened 2022-05-03
Add view-timeline support

kevers-google:


src/scroll-timeline-base.js
```diff
@@ -510,3 +510,161 @@ export class ScrollTimeline {
...
+
+  // Currently specced as fit with proposal to rename in order to more naturally
+  // support start and end transitions.
+  get range() {
+    return scrollTimelineOptions.get(this).range;
+  }
+
+  get currentTime() {
+    const unresolved = null;
+    if (this.phase === 'inacitve')
```
flackr, line 586: inacitve is mispelled

src/scroll-timeline-base.js
```diff
@@ -510,3 +510,161 @@ export class ScrollTimeline {
...
+      return unresolved;
+
+    // Compute the offset of the top-left corner of subject relative to
+    // top-left corner of the container.
+    const container = this.source;
+    const target = this.subject;
+
+    const targetBounds = target.getBoundingClientRect();
+    const containerBounds = container.getBoundingClientRect();
```
flackr, line 595: The BoundingClientRect of an overflow scroller is its visual borderbox rect. This means that top will be negative once the user has scrolled down which I think will result in an incorrect startOffset. The boundingClientRect also includes the border, but we should be measuring from the inside of the scroller. Here's a demo: scroll the scroller on https://jsbin.com/zimaxel/edit?html,css,js,output . This means at minimum I think you need to subtract the scrollTop from the containerBounds.

Maybe, however, we should be walking the offsetParent chain to the container and adding up the offsetTop / offsetLeft.

flackr (changes requested):

src/scroll-timeline-base.js
```diff
@@ -510,3 +510,161 @@ export class ScrollTimeline {
     return true;
   }
 }
+
+function getScrollParent(node) {
+  if (node === undefined)
+    return undefined;
+
+  if (node.scrollHeight > node.clientHeight ||
```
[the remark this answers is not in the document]
kevers-google: done.

src/scroll-timeline-base.js
kevers-google: Fixed.

src/scroll-timeline-base.js
```diff
@@ -510,3 +510,161 @@ export class ScrollTimeline {
...
+        (scrollerStyle.overflow == 'visible' ||
+         scrollerStyle.overflow == "clip")) {
+        return "inactive";
+    }
+
+    return "active";
+  }
+
+  // Currently specced as fit with proposal to rename in order to more naturally
+  // support start and end transitions.
```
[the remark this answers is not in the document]
kevers-google: done

src/scroll-timeline-base.js
```diff
@@ -510,3 +510,161 @@ export class ScrollTimeline {
...
+  // ViewTimelineOptions only has source. Furthermore, there is a strict
+  // relationship between subject and source (source is nearest scrollable
+  // ancestor of subject).
+
+  // Proceeding under the assumption that subject will be added to
+  // ViewTimelineOptions. Inferring the source from the subject if not
+  // explicitly set.
+  constructor(options) {
+    if (options.subject && !options.source) {
+      options.source = getScrollParent(options.subject.parentNode);
```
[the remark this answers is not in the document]
kevers-google: We rely on the source being set in order to attach the scroll event listener.  I can add a TODO to add a mutation observer to watch for style changes that would affect the resolution of 'source'.

src/scroll-timeline-base.js
```diff
@@ -510,3 +510,161 @@ export class ScrollTimeline {
...
+    }
+    super(options);
+
+    const details = scrollTimelineOptions.get(this);
+    details.subject = options && options.subject ? options.subject : undefined;
+    // TODO: what is the default range?
```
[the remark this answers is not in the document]

src/scroll-timeline-base.js
```diff
@@ -510,3 +510,161 @@ export class ScrollTimeline {
...
+    details.range = options && options.range ? options.range : 'contain';
+    // TODO: Handle insets.
+  }
+
+  get subject() {
+    return scrollTimelineOptions.get(this).subject;
+  }
+
+  // Proposing that we do away with before and after phase for a timeline and
+  // simply allow timeline time to extend outside the range of 0 to 100%.
```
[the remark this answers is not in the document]
kevers-google: Updated the comment.

src/scroll-timeline-base.js
kevers-google: Updated to do the walk.  You'll notice a slight tweak to the calculation for viewPos.

src/scroll-timeline-base.js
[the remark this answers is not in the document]
flackr: Adding a TODO is a good idea. We won't be able to observe style changes but we could at least re-evaluate the scroll source on a regular cadence. I have some ideas how to do this in a way that isn't too expensive - it will get slightly out of date but fix itself quickly.

src/scroll-timeline-base.js
```diff
@@ -512,18 +512,20 @@ export class ScrollTimeline {
...
 
-  if (node.scrollHeight > node.clientHeight ||
-      node.scrollWidth > node.clientWidth) {
-    return node;
-  } else {
-    return getScrollParent(node.parentNode);
-  }
+  const style = getComputedStyle(container);
+  let isScrollable = false;
+  style.overflow.split(" ").forEach(part => {
```
flackr, line 520: You don't need to check both overflow directions. Per https://www.w3.org/TR/css-overflow-3/#propdef-overflow-y overflow-x and overflow-y will both compute to auto or hidden respectively if the other direction is not visible or clip, so you can just check style.overflowY and it will implicitly include non-visible non-clip overflow from X.

src/scroll-timeline-base.js
```diff
@@ -512,18 +512,20 @@ export class ScrollTimeline {
...
+  if (!node)
     return undefined;
 
-  if (node.scrollHeight > node.clientHeight ||
-      node.scrollWidth > node.clientWidth) {
-    return node;
-  } else {
-    return getScrollParent(node.parentNode);
-  }
+  const style = getComputedStyle(container);
```
flackr, line 518: I think you meant node here, not container.

src/scroll-timeline-base.js
```diff
@@ -577,24 +590,33 @@ export class ViewTimeline extends ScrollTimeline {
...
-    const containerBounds = container.getBoundingClientRect();
-    const top = targetBounds.top - containerBounds.top;
-    const left = targetBounds.left - containerBounds.left;
+    // const targetBounds = target.getBoundingClientRect();
+    // const containerBounds = container.getBoundingClientRect();
+
+    let top = 0;
+    let left = 0;
+    let node = target;
+    while (node != container) {
```
flackr, line 615: Note, we should avoid an infinite loop here in case container is not actually an offsetParent of node. We should probably also throw an error when you attempt to use such a container.

src/scroll-timeline-base.js
```diff
@@ -577,24 +590,33 @@ export class ViewTimeline extends ScrollTimeline {
...
+      left += node.offsetLeft;
+      top += node.offsetTop;
+      node = node.parentNode;
```
flackr, line 618: s/parentNode/offsetParent

src/scroll-timeline-base.js
kevers-google: There should never be an infinite loop since a node cannot be its own offsetParent, but you could hit a null offsetParent.

Regarding throwing an exception, should this be specced?  It would seem a bit spammy to throw on each scroll position update.  Perhaps it should be caught during style update.  Alternatively, the phase should simply be inactive.  I'm implemented that later for now with a TODO to revisit. Also filed a spec issue.

src/scroll-timeline-base.js

src/scroll-timeline-base.js
kevers-google: done.

src/scroll-timeline-base.js
kevers-google: done.

src/scroll-timeline-base.js
```diff
@@ -510,3 +510,192 @@ export class ScrollTimeline {
...
+
+    if (container != document.scrollingElement &&
+        (scrollerStyle.overflow == 'visible' ||
+         scrollerStyle.overflow == "clip")) {
+        return "inactive";
+    }
+
+    // This check is not in the spec.
+    // http://github.com/w3c/csswg-drafts/issues/7259
+    // Update once specced.
```
flackr, line 592: nit: TODO

flackr (approved):

src/scroll-timeline-base.js

PR #59, opened 2022-07-06
Update list of demos

mehdi-kazemi:
Some of the demos use scroll-offset which has been removed from the spec, they should be updated once the view-timeline polyfill is updated based on the spec.

### DIFF
--- a/src/scroll-timeline-css-parser.js
+++ b/src/scroll-timeline-css-parser.js
@@ -210,6 +210,9 @@ export class StyleParser {
           // Add 1s as duration to fix this.
           if(timelineName || hasAnimationTimeline) {
             if(!this.hasDuration(shorthand)) {
+              // TODO: Should keep track of whether duration is artificial or not,
+              // so that we can later track that we need to update timing to
+              // properly see duration as 'auto' for the polyfill.
               rule.block.contents = rule.block.contents.replace(
                 "animation:",
                 "animation: 1s "


demo/basic/progress.html
```diff
@@ -0,0 +1,186 @@
...
+}
+
+#progress {
+  position: sticky;
+  top: 0px;
+  width: 100%;
+  height: 10px;
+  background-color: red;
+  animation-timeline: progressTimeline;
+  animation: progress 2s linear;
```
kevers-google, line 22: duration is not necessary.  It defaults to 'auto', which should then use the intrinsic iteration duration, which is calculated in order to fill the timeline duration.

kevers-google (changes requested):
Overall, it looks good. Just a few nits.

src/scroll-timeline-css-parser.js
```diff
@@ -204,12 +204,34 @@ export class StyleParser {
...
           }
+
+          // If there is no duration, animationstart will not happen,
+          // and polyfill will not work which is based on animationstart.
+          // Add 1s as duration to fix this.
+          if(timelineName || hasAnimationTimeline) {
+            if(!this.hasDuration(shorthand)) {
+              rule.block.contents = rule.block.contents.replace(
+                "animation:",
+                "animation: 1s "
```
kevers-google, line 215: Please add a comment that we should track whether the duration is artificial so that we can later  track that we need to update timing to properly see duration as 'auto' for the polyfill.

PR #60, opened 2022-07-18
Start end time

kevers-google:
Support specifying a start and end time in element.animate for a ViewTimeline.


src/scroll-timeline-base.js
```diff
@@ -132,155 +136,41 @@ function resolvePx(cssValue, resolvedLength) {
...
+function updateSource(timeline, source) {
+  const oldSource = scrollTimelineOptions.get(timeline).source;
+  if (oldSource == source)
+    return;
+
+  const listener = () => {
+    updateInternal(timeline);
+  };
+  if (oldSource)
+    scrollEventSource(oldSource).removeEventListener("scroll", listener);
```
flackr, line 170: I don't think this will work, isn't listener different in every call to updateSource? I think you'd need to keep the old bound listener.

src/scroll-timeline-base.js
```diff
@@ -515,6 +309,11 @@ function getScrollParent(node) {
   if (!node)
     return undefined;
 
+  if (!(node instanceof HTMLElement)) {
+     return node.parentNode ? getScrollParent(node.parentNode)
```
flackr, line 313: So, walking the parentNode chain is incorrect. We should be walking the containing block chain. E.g. a fixed position element's source should be `document.scrollingElement` in the absence of any fixed position containing ancestors, and similarly absolute position often escapes scrollers. This is a pre-existing issue but how about adding a TODO for this?

src/scroll-timeline-base.js
```diff
@@ -527,6 +326,122 @@ function getScrollParent(node) {
...
+
+    case 'exit':
+      // Range of scroll offsets where the subject element overlaps the
+      // logical-end edge of the viewport.
+      startOffset = viewPos;
+      endOffset = viewPos + viewSize;
+      break;
+  }
+
+  // TODO: Clamping of offsets is not specced. Update once ratified.
```
flackr, line 423: Link to the issue - also the current leading proposal in issue is not to clamp so perhaps we should follow that.

src/scroll-timeline-base.js
```diff
@@ -587,115 +505,22 @@ export class ViewTimeline extends ScrollTimeline {
         return "inactive";
```
flackr, line 505: I think you should be able to return super.phase to invoke the base class getter and not need to repeat the logic.

flackr (changes requested):

package.json
```diff
@@ -38,7 +38,7 @@
     "kleur": "^3.0.3",
     "microbundle": "^0.13.1",
     "run-all": "^1.0.1",
-    "saucelabs": "^7.1.3",
+    "saucelabs": "^7.1.2",
```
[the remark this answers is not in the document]
kevers-google: Vulnerability in a dependent library.  Picked up by npm audit.

src/scroll-timeline-base.js
```diff
@@ -85,13 +89,13 @@ export function calculateTargetEffectEnd(animation) {
...
  * @returns {Array} all currently installed parsers
  */
-export function installScrollOffsetExtension(parseFunction, evaluateFunction) {
-  extensionScrollOffsetFunctions.push({
-    parse: parseFunction,
-    evaluate: evaluateFunction,
-  });
-  return extensionScrollOffsetFunctions;
-}
+// export function installScrollOffsetExtension(parseFunction, evaluateFunction) {
```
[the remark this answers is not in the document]
kevers-google: Oops, meant to remove this rather than simply comment out.

src/scroll-timeline-base.js
kevers-google: Agreed. Would need to reference count to determine if we need to keep the old listener around since we could have multiple scroll-linked animations.  This logic was moved from ScrollTimeline's setter for source.  I can update to only remove if the animation list becomes empty.

src/scroll-timeline-base.js
kevers-google: Agreed. Written prior to removing the obsolete code from  ScrollTimeline.  I'll refactor.

src/scroll-timeline-base.js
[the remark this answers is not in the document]
kevers-google: done

src/proxy-animation.js
```diff
@@ -1542,6 +1589,57 @@ export class ProxyAnimation {
...
+
+  parts.forEach(part => {
+    if (part.endsWith('%'))
+      offsets.push(parseFloat(part));
+    else
+      names.push(part);
+  });
+
+  if (names.length > 2 || offsets.length > 2 || offsets.length == 1) {
+    // Reject value and use default.
```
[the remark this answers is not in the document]
kevers-google: Invalid / unparseable CSS input is typically ignored.  Though not CSS input when called from JavaScript, I think the same principle should apply. We can always revisit once resolved in the spec, as any throw would need to be clearly specced.

src/scroll-timeline-base.js

src/scroll-timeline-base.js
kevers-google: Added a TODO.

src/scroll-timeline-base.js
kevers-google: Added spec link.  Already landed code in Blink that clamps cover range.  I'd rather remain consistent with Blink implementation and update both sides if necessary once ratified on the spec.

src/proxy-animation.js
```diff
@@ -1553,6 +1651,10 @@ export function animate(keyframes, options) {
 
   if (timeline instanceof ScrollTimeline) {
     animation.pause();
+    if (timeline instanceof ViewTimeline) {
+      details = proxyAnimations.get(proxyAnimation);
```
bramus, line 1655: The `details` variable is not declared in this function. This will throw a `ReferenceError` in strict mode.

kevers-google: Fixed.

src/scroll-timeline-base.js
flackr: I don't think being consistent with Blink's implementation, when it doesn't match the spec is a goal. Using unclamped would allow my last comment to be a functional demo, whereas with clamped times it's not easy to emulate unclamped behavior.

src/proxy-animation.js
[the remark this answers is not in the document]
flackr: Yes, the CSS parser should ignore the error, but the JS API's typically throw errors when trying set invalid values through various JS animation methods, e.g. https://w3c.github.io/uievents/#event-types-list

src/scroll-timeline-base.js
kevers-google: Done.

src/proxy-animation.js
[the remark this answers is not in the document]
kevers-google: Done.

src/proxy-animation.js
```diff
@@ -382,16 +381,33 @@ function syncCurrentTime(details) {
...
   }
 }
 
+// Sets the time of the underlying animation, nudging the time slightly if at
+// a scroll-timeline boundary to remain in the active phase.
+function setNativeCurrentTime(details, time) {
+  const timeline = details.timeline;
+  const atScrollTimelineBoundary =
+      timeline.currentTime &&
+      timeline.currentTime.value == (this.playbackRate < 0 ? 0 : 100);
```
bramus, line 404: Getting an exception here when loaded as a module _(and thus in strict mode)_. Instead of `this.playbackRate` this should be `details.animation.playbackRate`.

Same with two lines down.

kevers-google: Fixed.  Added "use strict" to non-CSS demos.  This should make it easier to catch these errors.

src/proxy-animation.js
```diff
@@ -1623,8 +1624,7 @@ function parseTimeRange(value) {
   });
 
   if (names.length > 2 || offsets.length > 2 || offsets.length == 1) {
-    // Reject value and use default.
-    return timeRange;
+    throw new Error("Invalid time range");
```
flackr, line 1627: nit: This should be a TypeError, as we are unable to parse the supplied arguments as the expected type.

flackr (approved):
Looks good with one nit about the error type.

src/proxy-animation.js
kevers-google: done

PR #65, opened 2022-08-02
Get scroll parent

kevers-google:
Fix calculation the nearest ancestor scroll container.


src/scroll-timeline-base.js
```diff
@@ -296,25 +296,82 @@ export class ScrollTimeline {
...
+function getContainingBlock(element) {
+  switch (getComputedStyle(element).position) {
+    case 'static':
+    case 'relative':
+    case 'sticky':
+      return findClosestAncestor(element, isBlockContainer);
+
+    case 'absolute':
+    case 'fixed':
+      return findClosestAncestor(element, isAbsoluteOrFixedElementContainer);
```
flackr, line 354: This skips step 2:
> If the position property is absolute, the containing block is formed by the edge of the padding box of the nearest ancestor element that has a position value other than static (fixed, absolute, relative, or sticky).

tldr; if position absolute, you need to consider the above condition or any fixed container condition.

src/scroll-timeline-base.js
```diff
@@ -296,25 +296,82 @@ export class ScrollTimeline {
...
   switch(style['overflow-x']) {
     case 'auto':
     case 'scroll':
     case 'hidden':
-      return node;
+      return containingBlock;
 
     default:
-      return getScrollParent(node.parentNode);
+      return getScrollParent(containingBlock);
```
flackr, line 374: While technically correct, I do prefer looping over recursion as it's slightly more efficient.

src/scroll-timeline-base.js
```diff
@@ -350,28 +358,30 @@ function getContainingBlock(element) {
...
-    case 'auto':
-    case 'scroll':
-    case 'hidden':
-      return containingBlock;
-
-    default:
-      return getScrollParent(containingBlock);
+  while (true) {
+    const containingBlock = getContainingBlock(node);
+    if (!containingBlock)
```
flackr, line 374: nit: I think it's slightly cleaner to move this to the while condition, e.g.:

```js
while (node = getContainingBlock(node)) {
  const style = getComputedStyle(node);
  switch(style['overflow-x']) {
      case 'auto':
      case 'scroll':
      case 'hidden':
        return node;
  }
}
return document.scrollingElement;
```

flackr (approved):
LGTM with minor cleanup

src/scroll-timeline-base.js
kevers-google: done.  Also added missing special case rule for the body element.

src/scroll-timeline-base.js
```diff
@@ -369,20 +369,24 @@ function getScrollParent(node) {
...
       case 'scroll':
       case 'hidden':
-        return containingBlock;
+        // https://drafts.csswg.org/css-overflow-3/#overflow-propagation
+        // The UA must apply the overflow from the root element to the viewport;
+        // however, if the overflow is visible in both axis, then the overflow
+        // of the first visible child body is applied instead.
+        if (node.tagName != "BODY" ||
+            getComputedStyle(document.scrollingElement).overflow != "visible")
+          return node;
```
flackr, line 384: nit: I think it would be clearer if you flipped the condition and used this case for the exception (the body element in these special circumstances) to return document.scrollingElement and then left the return below as returning the node.

flackr (approved):
LGTM thanks Kevin!

PR #66, opened 2022-08-11
Add support for view-timeline (only name & axis)

mehdi-kazemi:
Parse view-timeline, view-timeline-name & view-timeline-axis
when view progress subject declares them, then create a new
ViewTimeline with the corresponding subject and axis,
and proxy the starting animation using this viewTimeline.


src/scroll-timeline-css.js
```diff
@@ -99,10 +112,16 @@ export function initCSSPolyfill() {
       const timelineName = parser.getScrollTimelineName(anim.animationName, evt.target);
       if (timelineName) {
         const scrollTimeline = createScrollTimeline(timelineName);
-        if (anim.timeline != scrollTimeline) {
+        const viewTimeline = createViewTimeline(timelineName);
```
kevers-google, line 115: This section is a bit confusing since a ViewTimeline is a ScrollTimeline. Suggest expanding the createScrollTimeline method to look at the options and then pick which timeline constructor to call based on source or subject being provided.  Downstream of timeline creation it doesn't matter if it is a generic or specialized ScrollTimeline so we can avoid the if/else logic, where both branches are identical in structure, differing only by the timeline variable.

src/scroll-timeline-css-parser.js
```diff
@@ -22,13 +25,16 @@ const ANIMATION_KEYWORDS = [
   'ease', 'linear', 'ease-in', 'ease-out', 'ease-in-out'
 ];
 
+const VIEW_TIMELINE_AXES = ['block', 'inline',  'vertical', 'horizontal'];
```
kevers-google, line 28: VIEW_TIMELINE_AXIS_(NAMES|TYPES)?

VIEW_TIMELINE_AXES and VIEW_TIMELINE_AXIS are close enough that it can make the code harder to read.

src/proxy-animation.js
```diff
@@ -751,7 +751,7 @@ function fractionalStartDelay(details) {
   if (!(details.timeline instanceof ViewTimeline))
     return 0;
 
-  const startTime = details.timeRange.start;
+  const startTime = details?.timeRange?.start || {name: 'cover', offset: CSS.percent(0)};
```
kevers-google, line 754: Suggest that we consolidate how we pick up the default for the start and end times for CSS and programmatic animations.  For programmatic, we set default values if not specified in the view timeline options.  This code block is essentially introducing a second place where we do "auto" to start and end time mapping.  

There is a case to be made for not filling in defaults in the element.animate path depending on decisions for getComputedTiming.

mehdi-kazemi: Added a TODO to consolidate after adding support for animation-delay & animation-end-delay.

src/scroll-timeline-css-parser.js
```diff
@@ -94,6 +100,16 @@ export class StyleParser {
     return null;
   }
 
+  getViewTimelineOptions(timelineName) {
+    // TODO: Take into account the scoping of the ViewTimelines
+    // https://github.com/w3c/csswg-drafts/issues/7047
+    for (let i = this.subjectSelectorToViewTimeline.length - 1; i >= 0; i--)
```
kevers-google, line 106: {} around body of for loop.

src/scroll-timeline-css-parser.js
```diff
@@ -206,6 +224,42 @@ export class StyleParser {
...
+    if(!hasViewTimeline && !hasViewTimelineName) return;
+
+    let viewTimeline = {selector: rule.selector, name: '', axis: 'block'};
+
+    if(hasViewTimeline) {
+      const parts = this.extractMatches(rule.block.contents, RegexMatcher.VIEW_TIMELINE, separator=' ');
+      if(parts.length == 1) {
+        viewTimeline.name = parts[0];
+      } else if(parts.length == 2) {
+        if(VIEW_TIMELINE_AXIS_TYPES.includes(parts[0])) {
```
kevers-google, line 241: optional: {} are not required for single line if-else.

src/scroll-timeline-css.js
```diff
@@ -72,16 +73,22 @@ function isDescendant(child, parent) {
...
-  return scrollTimeline;
+  if(options.source) {
+    const sourceElement = getSourceElement(options.source);
+    return new ScrollTimeline({
+      ...(sourceElement ? { source: getSourceElement(options.source) } : {}),
+      ...(options.orientation != "auto" ? { orientation: options.orientation } : {}),
+    });
+  } else {
+    return new ViewTimeline({
+      subject: document.querySelector(options.selector),
```
kevers-google, line 88: Please add TODO here.  This will catch the first matching selector, which may not be the correct one.  We want the last upstream match.

src/scroll-timeline-css-parser.js
```diff
@@ -94,6 +100,16 @@ export class StyleParser {
     return null;
   }
 
+  getViewTimelineOptions(timelineName) {
```
kevers-google, line 103: return { subject: element }

Then no need for extra downstream processing.  Simply call the new ViewTimeline.

src/scroll-timeline-css.js
```diff
@@ -72,16 +73,22 @@ function isDescendant(child, parent) {
...
   if (!options) return null;
 
-  const sourceElement = getSourceElement(options.source);
-
-  const scrollTimeline = new ScrollTimeline({
-    ...(sourceElement ? { source: getSourceElement(options.source) } : {}),
-    ...(options.orientation != "auto" ? { orientation: options.orientation } : {}),
-  });
-  return scrollTimeline;
+  if(options.source) {
```
kevers-google, line 80: We can probably do more upstream processing.  Then this block looks like:

return options.source ? new ScrollTimeline(options) : new ViewTimeline(options);

src/scroll-timeline-css-parser.js
```diff
@@ -94,6 +100,16 @@ export class StyleParser {
...
+    // TODO: Take into account the scoping of the ViewTimelines
+    // https://github.com/w3c/csswg-drafts/issues/7047
+    for (let i = this.subjectSelectorToViewTimeline.length - 1; i >= 0; i--)
+      if(this.subjectSelectorToViewTimeline[i].name == timelineName)
+        return this.subjectSelectorToViewTimeline[i];
```
kevers-google, line 108: target.getParentElement.querySelectorAll(...);  Then pick the last matching value upstream.

kevers-google:
LGTM

kevers-google (approved):
LGTM

PR #67, opened 2022-08-16
Parse animation-delay & animation-end-delay

mehdi-kazemi:
Save 'animation-delay' & 'animation-end-delay' alongside
'animation-timeline' & 'animation-name', all of which
are associated with the selector of a target that wants to
be animated. This will help us retrive the correct
Timeline when 'animationstart' happens.


src/proxy-animation.js
```diff
@@ -1594,20 +1590,18 @@ export class ProxyAnimation {
...
   const timeRange = {
-    start: {
-      name: 'cover',
-      offset: CSS.percent(0)
-    },
-    end: {
-      name: 'cover',
-      offset: CSS.percent(100)
-    }
+    start: { offset: CSS.percent(0) },
```
kevers-google, line 1599: Your change makes it mandatory to have at least one phase name.  As this is not clearly specced, I'm OK with this, but should be aware that this is a change in behavior.  The conditions for the throw should be updated accordingly.

We are still duplicating the start and end offsets.  I realize that this is to cover the case where a phase name is given without offsets.  I'm sure there is a cleaner way of doing this.

src/proxy-animation.js
```diff
@@ -818,7 +814,7 @@ export class ProxyAnimation {
       effect: null,
       // Range when using a view-timeline. The default range is cover 0% to
       // 100%.
-      timeRange: null,
+      timeRange: timeline instanceof ScrollTimeline ? timeline.timeRange : null,
```
kevers-google, line 817: timeRange cannot be a member of timeline since multiple animations can link to the same timeline with different ranges. Also we avoid hanging properties directly off of timeline that are not explicitly part of the API for encapsulation.  That's why we use the animation and timeline maps.  the map links the animation or timeline object to an object for storing additional information that is not exposed to the end user.

src/scroll-timeline-base.js
```diff
@@ -213,6 +213,7 @@ export class ScrollTimeline {
 
       // View timeline
       subject: null,
+      timeRange: null,
```
kevers-google, line 216: This is a property of the animation and not the timeline.

src/scroll-timeline-css-parser.js
```diff
@@ -127,17 +135,27 @@ export class StyleParser {
...
       const options = this.subjectSelectorToViewTimeline[i];
       if(options.name == timelineName) {
-        const allSubjects = target.parentElement.querySelectorAll(options.selector);
+        const allSubjects = document.querySelectorAll(options.selector);
         if(allSubjects.length) {
           return {
             subject: allSubjects[allSubjects.length - 1],
-            axis: options.axis
+            axis: options.axis,
+            timeRange: parseTimeRange(this.concatAnimationDelays(animOptions))
```
kevers-google, line 158: Multiple animations can share the same timeline but have different timeline ranges.

src/scroll-timeline-css-parser.js
```diff
@@ -300,24 +316,29 @@ export class StyleParser {
...
-          'animation-name': animationNames[i % animationNames.length],
-          'animation-timeline': timelineNames[i % timelineNames.length]
-        });
-      }
-    }
+    const hasAnimationDelay = rule.block.contents.includes("animation-delay:");
+    const hasAnimationEndDelay = rule.block.contents.includes("animation-end-delay:");
+
+    let animationDelays = [];
+    let animationEndDelays = [];
```
kevers-google, line 323: Note that endDelay though a property of animation-effect-timing does not map to a CSS longhand property name.  Probably deserving of a comment.  We'll need to update this in the css-animations-2 (?) spec.

src/proxy-animation.js
mehdi-kazemi: Updated, I didn't touch parseTimeRange() other than extracting default values in separate functions defaultAnimationDelay() & defaultAnimationEndDelay(). Instead I added another function parseAnimationDelays() that works with 'animation-delay' & 'animation-end-delay'

src/scroll-timeline-base.js
mehdi-kazemi: Removed,

src/proxy-animation.js
mehdi-kazemi: Updated, I pass timeRange to ProxyAnimation constructor and save it in proxyAnimations map.

src/proxy-animation.js
```diff
@@ -1596,7 +1598,8 @@ function parseOneAnimationDelay(delay, defaultOffset) {
 
   const parts = delay.split(' ');
 
-  if(parts.length == 2 && !parts[1].endsWith('%'))
+  if(!ANIMATION_DELAY_NAMES.includes(parts[0]) ||
+    (parts.length == 2 && !parts[1].endsWith('%')))
```
kevers-google, line 1602: Thanks for adding the additional validation.  We should also verify that the parseFloat call succeeds (didn't return NaN). Calling with 'cover rubbish%" will now fail with:

Uncaught TypeError: Failed to execute 'percent' on 'CSS': The provided double value is non-finite.

mehdi-kazemi: Done

src/proxy-animation.js
```diff
@@ -1602,7 +1602,15 @@ function parseOneAnimationDelay(delay, defaultOffset) {
     (parts.length == 2 && !parts[1].endsWith('%')))
     throw TypeError("Invalid animation delay");
 
-  const offset = parts.length == 2 ? CSS.percent(parseFloat(parts[1])) : defaultOffset;
+  let offset = defaultOffset;
+  if(parts.length == 2) {
+    const percentage = parseFloat(parts[1]);
+    if(Number.isNaN(percentage))
+      throw TypeError(`\"${parts[1]}\" is not a valid percentage for animation delay`);
+    else
```
kevers-google, line 1610: Don't need else due to the throw.

kevers-google (approved):

PR #68, opened 2022-08-18
Add support for animation-time-range

mehdi-kazemi:
animation-time-range is shorthand for animation-delay
and animation-end-delay.
Duplicate demo of 'animation-delay' & 'animation-end-delay'
to work with 'animation-time-range' instead.


src/proxy-animation.js
```diff
@@ -1619,6 +1619,9 @@ function defaultAnimationDelay() { return { name: 'cover', offset: CSS.percent(0
 function defaultAnimationEndDelay() { return { name: 'cover', offset: CSS.percent(100) }; }
 
 function parseAnimationDelays(animOptions) {
+  if(animOptions['animation-time-range'])
```
kevers-google, line 1622: What happens if you specify a time-range and animation-delay? Per spec, longhangs stomp shorthands, so we should have the end-delay calculated based on parsing animation-time-range by a start-delay based on parsing animation-delay.  Can you add a test to ensure that  this works.

PR #71, opened 2022-09-13
View timeline delay

kevers-google:
Adds support for delay and endDelay for programmatic animations on a view timeline.


src/proxy-animation.js
```diff
@@ -1680,6 +1680,32 @@ export function animate(keyframes, options) {
...
+        }
+        return null;
+     }
+  };
+
+  const updateDelay = (timelineOffset, value) => {
+    if (!value)
+      return;
+
+    // TODO(kevers): Update property names once ratified.
```
flackr, line 1698: nit: Link issue.

flackr (approved):
lgtm with nit

src/proxy-animation.js
kevers-google: done

PR #72, opened 2022-09-14
Add support for insets

mehdi-kazemi:


demo/view-timeline/index.html
```diff
@@ -110,7 +123,23 @@
     <div class="progress-bar-progress" style="top: 170px;"></div>
     <div class="progress-bar-progress" style="top: 220px;"></div>
     <div class="progress-bar-progress" style="top: 270px;"></div>
+    <div class="progress-bar-progress" style="top: 320px;"></div>
```
kevers-google, line 126: I'm a bit worried that this demo is getting too busy. 10 progress bars feels like a lot.  How about rename the current demo to indicate it's for view-timeline delays (timeRange was a placeholder and it appears we will use delay and endDelay instead).  For insets, we could have a nice visual with a box smaller than the viewport by the inset amount and show that the animated range aligns with the bounds of the inset box.

src/scroll-timeline-base.js
```diff
@@ -447,39 +448,39 @@ function range(timeline, phase) {
...
 
-  const scrollPos = directionAwareScrollOffset(container, orientation);
+  const inset = parseInset(details.inset, containerSize);
   let startOffset = undefined;
   let endOffset = undefined;
 
   switch(phase) {
     case 'cover':
       // Range of scroll offsets where the subject element intersects the
       // source's viewport.
```
kevers-google, line 458: These comments no longer align with the equation.  My preference is to change this to read "adjusted" viewport and update the viewPos and containerSize ahead of the switch to reflect the adjusted viewport.  Then all of the equations within the switch remain unchanged.

demo/view-timeline/with-inset.html
```diff
@@ -0,0 +1,118 @@
...
+  const createProgressAnimation = (bar, range, orientation, inset=null) => {
+    const subject = document.getElementById('subject');
+    const viewTimeline = new ViewTimeline({
+      'subject': subject,
+      'axis': orientation,
+      'inset': inset
+    });
+    bar.animate( { width: ['0px', '200px' ] }, {
+      timeline: viewTimeline,
+      timeRange: `${range}`,
```
kevers-google, line 103: Suggest using delay and endDelay since timeRange is likely to be deprecated.  The current proposal is to change animation-delay to be a shorthand property which is start delay if one item is specified and start + end if two values are specified.

src/scroll-timeline-base.js
```diff
@@ -489,6 +491,38 @@ function range(timeline, phase) {
...
+  if(!value)
+    return inset;
+
+  const parts = value.split(' ');
+  const insetParts = [];
+  parts.forEach(part => {
+    if(part.endsWith("%"))
+      insetParts.push(containerSize / 100 * parseFloat(part));
+    else if(part.endsWith("px"))
+      insetParts.push(parseFloat(part));
```
kevers-google, line 506: Add TODO for supporting relative lengths (e.g. em).

kevers-google (approved):

PR #77, opened 2022-09-22
Ignore comments when parsing CSS

mehdi-kazemi:


src/scroll-timeline-css-parser.js
```diff
@@ -415,7 +415,7 @@ export class StyleParser {
 
   eatComment(p) {
     this.assertString(p, "/*");
-    this.eatUntil("*/", p);
+    this.eatUntil("*/", p, true);
```
kevers-google, line 418: Does the container query polyfill have the same bug?  If so, please file an issue.

demo/view-timeline-css/index.html
```diff
@@ -23,8 +23,23 @@
   width: 50px;
   height: 50px;
   animation-timeline: foo;
+  /* animation-time-range: enter 0% 100%; */
   animation: linear anim;
 }
+
+/*
```
kevers-google, line 30: Add a comment that this is deliberately commented out.  Otherwise it could be deleted as part of a cleanup.  Here and above.

src/scroll-timeline-css-parser.js
mehdi-kazemi: Will take care of it.

kevers-google (approved):

Issue #78, opened 2022-09-23
Support styles outside of style tag


jh3y:
Currently using the polyfill with CSS styles only works if the styles are inlined into the `head`:

```html
<style>
  .trigger { view-timeline-name: --trigger; }
  .el { animation-timeline: --trigger; }
</style>
```

If you put those styles in a file like `style.css` and link them with:

```html
<link rel="stylesheet" href="style.css" />
```

The polyfill fails silently because it doesn't consider loaded styles.

Until that's supported, might it be worth adding a `console.warn` that any styles not inlined won't be detected? Happy to open a PR for this.

@jh3y

AndrewKGuan:
Does it require all styles to be inline or just the ones that are related to the animating elements?  But either way, may need support to external stylesheets for larger scaled sites to make this polyfill useable.

bramus:
Just the scroll-animation related ones.

clementmas:
You should include this crucial info in the readme.

webGeniusDK:
I have all animation-timeline related css inside style tags in the head of my page (lots of other external css), but I still get the warning: "Non-Inline StyleSheets detected: ScrollTimeline polyfill currently only supports inline styles within style tags"

bramus:
That’s expected behavior (for now): the polyfill detects that you have external stylesheets, so warns you about it potentially not working. Awaiting actual support for parsing of external stylesheets, the wording could use a minor tweak to make it more clear that things could _potentially_ not work.

webGeniusDK:
Ok.. Thanks. So you're saying that the polyfill might not work at this moment, if the document has other external stylesheets, with styles not related to this at all? (Just need to understand. I am very grateful for you work)

webGeniusDK:
Ok... removed all animation related styles in external stylesheets, and still it doesn't work. Just writing this so you know it. I will wait untill your work is further ahead. Thanks again!

calinoracation:
I've started a super naive exploration into experimentally parsing the external stylesheets. Can't see it's actually working on my test yet, it parses much of it, but isn't exactly fully working yet.

```typescript
function handleLinkedStylesheet(sheet: HTMLLinkElement) {
  fetch(sheet.href).then(async (response) => {
    const newSheet = new CSSStyleSheet();
    const result = await response.text();
    let newSrc = parser.transpileStyleSheet(result, true);
    newSrc = parser.transpileStyleSheet(newSrc, false);
[...]

PR #81, opened 2022-09-23
feat: add console warning for non-inline styles

jh3y:
Interim solution for #78 to let the user know that non-inline styles are not supported 👍


src/index.js
```diff
@@ -25,6 +25,10 @@ import { initCSSPolyfill } from "./scroll-timeline-css"
 
 initCSSPolyfill();
 
+if ([...document.styleSheets].filter(s => s.href !== null).length) {
+  console.warn("Non-Inline StyleSheets detected: ScrollTimeline polyfill currently only supports inline styles within style tags");
```
kevers-google, line 29: Nit: 80-column character limit.
console warn("...." +
                       "..."); 

to avoid a long line.

kevers-google (approved):

src/index.js
jh3y: That's fair! Lemme change that quick 👍

jh3y:
Resolved 🔧  Think it jus' needs a re-approval on the workflow 🙏

kevers-google (approved):

PR #83, opened 2022-10-04
Take scoping of the ViewTimelines into account

### LINKED ISSUE #82: CSS View Timelines are shared?
I’ve [created a CoverFlow demo](https://codepen.io/bramus/pen/MWVExQg?editors=1010) which approaches the problem using the JS side of things. Each `<li>` is linked to a `ViewTimeline`, and during its `contain` phase animations are running. This works fine.

https://user-images.githubusercontent.com/213073/193609107-7480d07d-42d8-4103-aaf5-06584c53d6b7.mp4

When trying to build [a CSS version of it](https://codepen.io/bramus/pen/GRdGoKy/eb996da60fd733dab962105bbe8d0bee), it doesn’t seem to work: all animations on all `<li>` elements only run when the very last `<li>` of the bunch comes into view. It looks like the timeline is shared between all instances, whereas it is expected for each `<li>` to have their own instance attached.

https://user-images.githubusercontent.com/213073/193609185-8713aeea-0be0-4660-8c0d-aa50322411d1.mp4

Used Code:

```css
.cards li {
	view-timeline-name: li-in-and-out-of-view;
	view-timeline-axis: inline;

	animation: 1s linear adjust-z-index;
	animation-timeline: li-in-and-out-of-view;
	animation-time-range: contain 0% 100%;
}

.cards li > img {
	animation: 1s linear rotate-cover;
	animation-timeline: li-in-and-out-of-view;
	animation-time-range: contain 0% 100%;
}
```

_(Used property names follow those the current version of the polyfill supports, not of the latest spec)_

bramus:
In https://github.com/w3c/csswg-drafts/issues/7047 we resolved that timeline search looks at preceding siblings and ancestors, recursively. [`Element.closest()` does exactly that](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest), so we can use that to determine the subject.

Fixes #82


bramus:
To verify the behavior of this PR, I have updated [the CSS demo](https://codepen.io/bramus/pen/GRdGoKy/eb996da60fd733dab962105bbe8d0bee) mentioned #82 to load the JS file from commit ccfcac4ce57541fb5be66a08d978bf9448a185b4 from this PR:

Before, without the fix:

```html
<script src="https://flackr.github.io/scroll-timeline/dist/scroll-timeline.js"></script>
```

After, with the fix:

```html
<script src="https://rawcdn.githack.com/flackr/scroll-timeline/ccfcac4ce57541fb5be66a08d978bf9448a185b4/dist/scroll-timeline.js"></script>
```

src/scroll-timeline-css-parser.js
```diff
@@ -135,16 +135,14 @@ export class StyleParser {
...
-  getViewTimelineOptions(timelineName) {
-    // TODO: Take into account the scoping of the ViewTimelines
-    // https://github.com/w3c/csswg-drafts/issues/7047
+  getViewTimelineOptions(timelineName, target) {
     for (let i = this.subjectSelectorToViewTimeline.length - 1; i >= 0; i--) {
       const options = this.subjectSelectorToViewTimeline[i];
       if(options.name == timelineName) {
-        const allSubjects = document.querySelectorAll(options.selector);
-        if(allSubjects.length) {
+        const subject = target.closest(options.selector);
```
kevers-google, line 142: https://drafts.csswg.org/scroll-animations-1/#timeline-scope

My reading of "closest" is that it is an inclusive ancestor and does not consider upstream slblings.  Let me take a quick read of the Blink implementation to see how we dealt with this.

src/scroll-timeline-css-parser.js
```diff
@@ -135,16 +135,36 @@ export class StyleParser {
...
+    while (prevSibling) {
+      if (prevSibling.matches(selector)) {
+        return prevSibling;
+      }
+      prevSibling = prevSibling.previousElementSibling;
+    }
+
+    // Move up to parent
+    if (target.parentElement) {
+      return this.findPreviousSiblingOrAncestorMatchingSelector(target.parentElement, selector);
```
kevers-google, line 153: Nit: iteration is more efficient than recursion.
```
let candidate = target;
while (candidate) {
   if (candidate.matches(selector)) 
     return candidate;
   candidate = candidate.previousElementSibling || candidate.parentElement;
}
```

src/scroll-timeline-css-parser.js
bramus: Fixed in c98c9e2e98f832376de77204f2cd8df8519a0901

Standalone version of that function to (manually) test things: https://codepen.io/bramus/pen/RwyJOJM/1f63d147250c7f682be221b71f0e1961

src/scroll-timeline-css-parser.js
bramus: Clever! Addressed in 31b3d9585daa88455b69d170a16b5e7d68aad6d9

src/scroll-timeline-css-parser.js
```diff
@@ -136,24 +136,13 @@ export class StyleParser {
...
-      if (prevSibling.matches(selector)) {
-        return prevSibling;
-      }
-      prevSibling = prevSibling.previousElementSibling;
-    }
-
-    // Move up to parent
-    if (target.parentElement) {
-      return this.findPreviousSiblingOrAncestorMatchingSelector(target.parentElement, selector);
+    let candidate = target;
```
kevers-google, line 139: Optional nit: OK to leave the comments.  

// Target self ... 
// Walk previous sibling, then the parent.
// No match

are all useful comments and make the code easier to read.

kevers-google (approved):

src/scroll-timeline-css-parser.js
bramus: :)

bramus: Addressed in 48b223c56dab2c31bd689e967c8d6a0a33b33224

PR #84, opened 2022-10-04
Performance: Cache ProxyAnimation instances

bramus:
Currently, the CSS polyfill creates new `ProxyAnimation` instances every time an animation starts. As you’re scrolling back and forth, you can see performance stutter.

This PR caches created instances onto the targeted elements, which improves performance.

Tested using the demo from https://codepen.io/bramus/pen/GRdGoKy, the performance timelines look like this:

Before             |  After
:-------------------------:|:-------------------------:
  |


bramus:
FYI: Rebased this branch onto `master` to make sure it can be merged easily.

src/scroll-timeline-css.js
```diff
@@ -148,11 +148,28 @@ export function initCSSPolyfill() {
   // because we may lose some of the 'animationstart' events by the time 'load' is completed.
   window.addEventListener('animationstart', (evt) => {
     evt.target.getAnimations().filter(anim => anim.animationName === evt.animationName).forEach(anim => {
-      const result = createScrollTimeline(anim, anim.animationName, evt.target);
-      if (result.timeline && anim.timeline != result.timeline) {
-        const proxyAnimation = new ProxyAnimation(anim, result.timeline, result.animOptions);
+      // Create a cache on the element, to store all Proxy Animations in
+      if (!evt.target.proxiedAnimations) {
+        evt.target.proxiedAnimations = new Map();
```
kevers-google, line 153: We should not hang new stuff off element as it breaks encapsulation.  We have animation and timeline details to avoid exposing polyfill internals to the outside world.  

Instead we can add a global element map to the polyfill, i.e.:

let elementAnimationsMap = new WeakMap();

and then retrieve the map of proxied animations from there.  Please take a look at how  proxyAnimations.get(...) is used to retrieve animation details without exposing to the "outside world".  We should follow the same pattern here.

kevers-google (approved):

src/scroll-timeline-css.js
bramus: Addressed in 3fb2b7f03af35061c9b272ce2770c2feb4069839.

Note I could not use a `WeakMap` for the nested cache, as WeakMaps require Objects as their key. As the used `anim` always is a new Object, this didn’t work – it only created more Objects.

Using WeakMap             |  Using Map
:-------------------------:|:-------------------------:
image |  image

src/scroll-timeline-css.js
```diff
@@ -144,13 +144,34 @@ export function initCSSPolyfill() {
...
-      if (result.timeline && anim.timeline != result.timeline) {
-        const proxyAnimation = new ProxyAnimation(anim, result.timeline, result.animOptions);
+      // Create a per-element cache
+      if (!proxyAnimations.has(evt.target)) {
+        proxyAnimations.set(evt.target, new Map());
+      }
+      const elementProxyAnimations = proxyAnimations.get(evt.target);
+
+      // Store Proxy Animation in the cache
+      if (!elementProxyAnimations.has(anim.animationName)) {
```
flackr, line 161: `animation-name` allows listing the same animation more than once, e.g.
```css
#target {
  animation-name: spin, spin;
}
```
This is expected to produce 2 animations with the latter having a higher composite order, but I believe this change will result in a single animation.

kevers-google: We'll need to track the index as well as the name.

flackr: I created an alternative version in https://github.com/flackr/scroll-timeline/pull/89 which also implements getAnimations support in this polyfill.

PR #85, opened 2022-10-04
Fix feature detection

bramus:
This PR splits up the feature detection used in the CSS parser in two parts:

1. Separately detect support for `scroll-timeline` _(by checking for `animation-timeline: scroll()` support)_
2. Separately detect support for `view-timeline`

This will benefit browsers that land partial support, should that ever be the case.

Note that this doesn’t work in today’s Chrome Canary with the Experimental Web Platform Features enabled, as it never seems to trigger an `animationstart` event for animations that are linked to a `view-timeline`.


bramus:
FYI: Rebased this branch onto `master` to make sure it can be merged easily.

bramus:
> Do we still need this? I'm curious are there any browsers planning to ship one of scroll-timeline or view-timeline but not both?

Yeah, good question. In theory it is possible, so I think it’s better to err on the safe side of things here.

bramus:
Abandoning this PR after all. Giving all this some more thought, it’s highly unlikely that a vendor would implement STL but not VTL.

(The more obvious split they _could_ make is CSS vs JS,but that’s a different story)

PR #86, opened 2022-10-04
Add support for phase linked offsets in CSS

mehdi-kazemi:


src/scroll-timeline-css-parser.js
```diff
@@ -362,19 +365,110 @@ export class StyleParser {
...
+      enter 100% { opacity: 1 }
+      exit 0% { opacity: 1 }
+      exit 100% { opacity: 0 }
+    }
+  with this:
+    {
+      0% { opacity: 0 }
+      1% { opacity: 1 }
+      2% { opacity: 1 }
+      3% { opacity: 0 }
```
kevers-google, line 391: What if you have the following 

{ 
  0%: ...
  enter 0%: ...
  1%: ...
}

I see from the implementation below that even non-timeline offset entries are remapped if any require remapping, but this is not clear from the comment.  

It should also be made clear in the comment that the mapping will get further stomped downstream as 1%, 2%, etc are unlikely to be the correct offsets. You mention creating a mapping, but don't say how it will be used.

Though it seems crazy to have more than 100 keyframes, it is possible. So instead of incrementing by 1%, it may make sense to increment by i/N * 100% where N is the number of frames.

kevers-google (approved):
LGTM

PR #87, opened 2022-10-05
Add support for Anonymous Scroll Timeline

mehdi-kazemi:


demo/view-timeline-css/with-named-timeline-range-keyframe-selectors.html
```diff
@@ -0,0 +1,57 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <title>Named Timeline Range Keyframe Selectors</title>
```
kevers-google, line 6: Demo does not look to be related to the rest of the patch.  Was this demo intended for the previous patch on timeline offsets in keyframes?

demo/basic/anonymous-scroll-progress-timeline.html
```diff
@@ -0,0 +1,42 @@
+<!DOCTYPE HTML>
+<title>Anonymous Scroll Progress Timeline</title>
```
kevers-google, line 2: Not linked to the demo list page.

src/scroll-timeline-css-parser.js
```diff
@@ -123,7 +126,23 @@ export class StyleParser {
       return null;
   }
 
-  getScrollTimelineOptions(timelineName) {
+  getAnonymousScrollTimelineOptions(timelineName, target) {
+    const options = this.anonymousScrollTimelineOptions.get(timelineName);
+    if(options) {
+      return {
+        source: (options.source == 'root' ? document.scrollingElement : getScrollParent(target)),
```
kevers-google, line 133: By resolving immediately, we won't update the source in response to DOM changes. We have WPT tests for this already for the JS API.  How about making source Element | string/enum and then resolving a non-element source at the same time we validate/update the view timeline source.

bramus:
Thanks for adding this @mehdi-kazemi! I do notice the polyfill pauses the animations just a tad too late?

If I load https://cdpn.io/pen/debug/KKoGNye it wil sometimes show the entire scrollbar on load _(it should be `scaleX(0)`)_, and often it shows a very little bit of the scrollbar. Once you start scrolling, the animation corrects itself.

Maybe this is food for a follow-up issue, in order to not block this PR.

src/scroll-timeline-css-parser.js
```diff
@@ -354,6 +375,48 @@ export class StyleParser {
...
   }
 
+  extractScrollTimelineNames(contents) {
+    const value = RegexMatcher.ANIMATION_TIMELINE.exec(contents)[1].trim();
+    const timelineNames = [];
+
+    value.split(",").map(part => part.trim()).forEach(part => {
+      if(isAnonymousScrollTimeline(part)) {
+        const name = this.saveAnonymousTimelineName(part);
+        timelineNames.push(name);
```
kevers-google, line 385: Possible for there to be a name collision between the auto-generated name for an anonymous scroll timeline and a named timeline.

mehdi-kazemi: That's true, I added ":" to the beginning of anonymous timeline names to get rid of the collision situation. 
':' is not allowed to be used in identifiers, even after obfuscation.
animation-timeline: foo, scroll(), :bar;     
// ":bar" is not valid here 

I've stashed adding ":true", ":false" to the end of names, please let me know if adding ':' to the beginning of anonymous timeline names doesn't work, then I'll upload the other approach.

src/scroll-timeline-css-parser.js
```diff
@@ -354,6 +375,50 @@ export class StyleParser {
...
+      }
+    });
+
+    return timelineNames;
+  }
+
+  saveAnonymousTimelineName(part) {
+    // Anonymous scroll timelines are given a name that starts with ':' to
+    // prevent collision with named scroll timelines.
+    const name = `:t${this.anonymousScrollTimelineOptions.size}`;
```
kevers-google, line 397: let nextAnonymousTimelineNameIndex = 0;

saveAnonymousTimelineName(part) {
   const name = `:t${nextAnonymousTimelineNameIndex++}`;
   ...
}

Presently, we have a "memory leak" in that we don't release the name / timeline association when the timeline is no longer needed.  Once we fix the leak, we then introduce the possibility of reusing the name since ...Options.size could shrink and then grow again.

src/scroll-timeline-css-parser.js
kevers-google: : prefix looks good for disambiguating.

src/scroll-timeline-css-parser.js
```diff
@@ -354,6 +376,50 @@ export class StyleParser {
...
+    // Anonymous scroll timelines are given a name that starts with ':' to
+    // prevent collision with named scroll timelines.
+    const name = `:t${this.nextAnonymousTimelineNameIndex++}`;
+    this.anonymousScrollTimelineOptions.set(name, this.parseAnonymousTimeline(part));
+    return name;
+  }
+
+  parseAnonymousTimeline(part) {
+    const openIndex = part.indexOf("(");
+    const closeIndex = part.indexOf(")");
```
kevers-google, line 405: OK with this for now, but we should consider rewriting as this implementation is not strict. Off the top of my head, I cannot think of valid CSS that would be misinterpreted; however, we could erroneously flag invalid CSS as valid. For strictness, we should likely use a regex with capture groups.

PR #92, opened 2022-10-25
Do not throw away keyframes mapping

bramus:
I had an issue where https://codepen.io/bramus/pen/VwdZGYZ did not work correctly. The keyframes used are these:

```css
@keyframes fade-in-on-enter--fade-out-on-exit {
	enter 0% {
		opacity: 0;
		transform: translateY(100%);
	}
	enter 100% {
		opacity: 1;
		transform: translateY(0);
	}
	exit 0% {
		opacity: 1;
		transform: translateY(0);
	}
	exit 100% {
		opacity: 0;
		transform: translateY(-100%);
	}
}
```

Each `<li>` in the demo has these keyframes applied, linked to a `view-timeline`. Only the first `<li>` animated out correctly, the other items did not use the correct keyframes.

_(You can test this in the demo by commenting out the loading of the forked polyfill and uncommenting the original one)_

While the adjustment of the keyframes to a mapping went correct, I did notice that only 1 animation got their keyframes updated in [`updateKeyFramesIfNecessary`](https://github.com/flackr/scroll-timeline/blob/09064fec2981d662bcebba7c4c8704719cfae896/src/scroll-timeline-css.js#L87), the rest did not. The culprit was [this line](https://github.com/flackr/scroll-timeline/blob/09064fec2981d662bcebba7c4c8704719cfae896/src/scroll-timeline-css.js#L135), which removes the mapping after the first element got its keyframes updated.

```js
parser.keyframeNamesSelectors.set(anim.animationName, null);
```

Because of that, successive elements with the same keyframes would never receive updated keyframes.

Removing this one line got my demo fixed. Other demos tested remained working fine with this change.


bramus:
Also did an alternative approach that caches the `sortedKeyFrames`: https://github.com/flackr/scroll-timeline/compare/master...bramus:scroll-timeline:fix-multiple-viewtimelines-with-same-keyframes-cache. 

After chatting with @mehdi-kazemi this was considered not feasible, as the `sortedKeyFrames` are dependent on offsets calculated from the `container`, `subject`, etc. which can differ.

mehdi-kazemi (approved):

PR #94, opened 2022-10-26
Add support for view-timeline-inset in CSS

mehdi-kazemi:


src/scroll-timeline-css-parser.js
```diff
@@ -314,35 +315,62 @@ export class StyleParser {
...
-        if(TIMELINE_AXIS_TYPES.includes(parts[0]))
-          viewTimeline.axis = parts[0], viewTimeline.name = parts[1];
-        else
-          viewTimeline.axis = parts[1], viewTimeline.name = parts[0];
+      const viewTimelines = this.extractMatches(rule.block.contents, RegexMatcher.VIEW_TIMELINE);
+      for(let tl of viewTimelines) {
+        parts = this.split(tl);
+        let options = {selector: rule.selector, name: '', axis: 'block', inset: null};
+        if(parts.length == 1) {
+          options.name = parts[0];
```
kevers-google, line 330: Don't we still need to set the axis to block in this case?

kevers-google: Confirmed offline that we are setting the default correctly in the view timeline constructor.

kevers-google (approved):

Issue #108, opened 2023-05-12
Breaks "TypeIt" animations


dream-ellie:
**TL;DR**
When this polyfill is imported, it breaks page animations that use TypeIt (https://github.com/alexmacarthur/typeit)

**Details**
There is a TypeIt animation on the page that works just fine until we add the polyfill in the <script> tag to the html page.

```
  <script src="https://flackr.github.io/scroll-timeline/dist/scroll-timeline.js" defer></script>
  <script src="https://unpkg.com/typeit@8.7.1/dist/index.umd.js" defer></script>
```
The animation disappears and doesn't work any more.
The same happens if we import the polyfill using "import" in a JS module file.

In my own setup I get the following exception in the browser console:

```
Uncaught (in promise) TypeError: Cannot set property id of #<e> which has only a getter
    at D (index.umd.js:2:2778)
    at R (index.umd.js:2:4767)
    at Object.go (index.umd.js:2:7725)
    at type.js:27:4
```
Digging into the polyfill code I found that the proxyAnimation class does not expose a setter for the `id`:
https://github.com/flackr/scroll-timeline/blob/29b7b6e1be4bb40192f987d7c001f6e2910e0912/src/proxy-animation.js#L1490-L1492

And it looks like the code that throws, is trying to set it.
```
      , C = e=>e?.getAnimations().find((t=>t.id === e.dataset.tiAnimationId))
      , D = ({cursor: e, frames: t, options: n})=>{
        let r = e.animate(t, n);
        return r.pause(),
        r.id = e.dataset.tiAnimationId,  // <===== THIS IS THE LINE THAT THROWS
        L((()=>{
            L((()=>{
                r.play()
            }
            ))
        }
        )),
        r
```

As I urgently want to fix it, will a PR with the following patch do the trick?

```diff
+  set id(value) {
+    proxyAnimations.get(this).animation.id = value;
+  }
```

Let me know if that will do it and I can submit a PR asap.

Here is a sandbox project to play with: https://codesandbox.io/s/interesting-buck-yyev82?file=/index.html

PR #109, opened 2023-05-12
Add 'id' setter to ProxyAnimation to fix broken TypeIt animations

dream-ellie:
**WHAT**
Added a setter to "id" attribute for ProxyAnimation class.

**WHY**
Fix https://github.com/flackr/scroll-timeline/issues/108 
When polyfill is loaded, it breaks TypeIt animations on the page

**HOW**
TypIt animations attempt to set animation ids. Missing setter was causing an exception.

**TEST**
I built it locally and tested with my own TypeIt setup. It solved the issue and everything is working fine for me.

@kevers-google @flackr: I see there is a plan to rewrite the ProxyAnimation class to be a "full-fledged proxy" so the issues like this one won't happen but this task seems ambitious and there are not enough hours in a day. But I need a fix in a very short time frame.

https://github.com/flackr/scroll-timeline/blob/29b7b6e1be4bb40192f987d7c001f6e2910e0912/src/proxy-animation.js#L769-L771

So, please could we have it merged meanwhile?


dream-ellie:
Hi @kevers-google @flackr may I please ask for someone to review this PR?
This is a pretty simple and small fix that solves the problem I raised in https://github.com/flackr/scroll-timeline/issues/108

dream-ellie:
> @dream-elite
> 
> Sorry for the slow response.
> 
> Can you please run "npm run dev" to generate the binaries. These are updated alongside the source and should be part of the pull request.

@kevers-google  Done

kevers-google (approved):

PR #113, opened 2023-06-05
Add support for win32 development

egirard:
Adding some windows (batch or command shell) support for build process.

### DIFF
--- a/src/scroll-timeline-base.js
+++ b/src/scroll-timeline-base.js
@@ -44,12 +44,12 @@ function updateInternal(scrollTimelineInstance) {
 
 /**
  * Calculates a scroll offset that corrects for writing modes, text direction
- * and a logical orientation.
+ * and a logical axis.
  * @param scrollTimeline {ScrollTimeline}
- * @param orientation {String}
+ * @param axis {String}
  * @returns {Number}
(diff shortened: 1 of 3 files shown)


src/scroll-timeline-base.js
```diff
@@ -589,9 +589,9 @@ export class ViewTimeline extends ScrollTimeline {
   // explicitly set.
   constructor(options) {
     if (options.axis) {
```
kevers-google, line 591: Can get rid of this special handling of axis.  ScrollTimeline and ViewTimeline both use axis instead of orientation now, and axis is already handled in the base class.

src/scroll-timeline-base.js
```diff
@@ -615,15 +615,15 @@ export class ViewTimeline extends ScrollTimeline {
     return scrollTimelineOptions.get(this).subject;
   }
 
-  // The orientation is called "axis" for a view timeline.
-  // Internally we still call it orientation.
+  // The axis is called "axis" for a view timeline.
+  // Internally we still call it axis.
   get axis() {
```
kevers-google, line 620: Remove.  Handled in ScrollTimeline.

PR #114, opened 2023-06-25
work with chrome canary now

Gaubee:
follow the [drafts](https://drafts.csswg.org/scroll-animations-1/)


Gaubee:
By the way, should open source projects unify code formatting tools?

src/scroll-timeline-css.js
```diff
@@ -140,8 +140,8 @@ function updateKeyframesIfNecessary(anim, options) {
 
 export function initCSSPolyfill() {
   // Don't load if browser claims support
-  if (CSS.supports("animation-timeline: works")) {
-    return;
+  if (CSS.supports("animation-timeline: none")) {
```
kevers-google, line 143: "animation-timeline: --works" is probably a better choice. Strictly speaking a null timeline can be supported even without scroll-driven animation support.

test/tests.config.json
```diff
@@ -18,7 +18,7 @@
     "/scroll-animations/css/at-scroll-timeline-cascade.html",
     "/scroll-animations/css/at-scroll-timeline-element-offsets.html",
     "/scroll-animations/css/at-scroll-timeline-inactive-phase.html",
-    "/scroll-animations/css/at-scroll-timeline-orientation.html",
+    "/scroll-animations/css/at-scroll-timeline-axis.html",
```
kevers-google, line 21: This file should be synced from WPT.  None of the .../css/at-scroll-* tests exist any longer.

kevers-google (changes requested):
Sorry to take so long to get back to you.  Overall it looks good.  Just a couple of nits.

Sadly, the polyfill is quite out of date now, so many more changes will be needed to get it back up and running.  One of the dangers of implementing a polyfill based on a spec that was still in flux.  If you are still eager to contribute, that's great.  Hope to get this back up and running soon.

test/tests.config.json
Gaubee: I will make RP for WPT.

kevers-google: What I'm getting at is that all of the at-scroll-timeline-* tests have been removed from wpt as they were testing obsolete syntax.  So we shouldn't update to exclude a non-existent test, but rather ensure that the json file reflects tests that are actually in wpt.  Also, updating test.config.json affects anyone testing this polyfill.  It may make sense to keep your own modified version of the config file to restrict testing to certain tests, but we'll want test coverage even for those that fail.  I believe these exclusions were added at a time when the spec was in flux and we were phasing out some old syntax.

test/tests.config.json
```diff
@@ -9,21 +9,26 @@
...
-    "/scroll-animations/css/at-scroll-timeline-cascade.html",
-    "/scroll-animations/css/at-scroll-timeline-element-offsets.html",
-    "/scroll-animations/css/at-scroll-timeline-inactive-phase.html",
-    "/scroll-animations/css/at-scroll-timeline-orientation.html",
-    "/scroll-animations/css/at-scroll-timeline-sampling.html",
-    "/scroll-animations/css/at-scroll-timeline-source.html",
-    "/scroll-animations/css/at-scroll-timeline-start-end.html",
-    "/scroll-animations/css/at-scroll-timeline-time-range.html",
-    "/scroll-animations/css/scroll-timeline-cssom.tentative.html"
+    "/scroll-animations/css/**/*.html",
```
kevers-google, line 12: Please do not augment the test exclusions. We have other developers working on the Polyfill as well.  Fine to keep this as your own exclusions file, but this should not be part of the pull request.

test/webdriver/index.js
```diff
@@ -12,6 +12,7 @@ const TEST_CONFIGS = require("../tests.config.json");
 const harnessTests = require("./harness-tests");
 const { resolve } = require("path");
 const os = require('os');
+const glob = require('tiny-glob');
```
kevers-google, line 15: Suggest pulling out the webdriver changes into a separate pull request as this work is unrelated to the orientation->axis rename.

kevers-google: Also note, that dist has been pulled out now that the binaries are automatically built and deployed by a script.

Gaubee:
I'm sorry that I don't have enough time to contribute to this project, I will close this PR first, thank you for your time.

egirard:
Thanks for your contribution, Gaubee - I'm going to clean up and reland your patch.

On a side-note, our team has cleaned up the infrastucture to make work like this easier - your patch helped us better understand those infra-related shortcomings.  So thanks for that as well.

PR #117, opened 2023-07-18
Support running wpt tests directly using --inject-script.

flackr:
This uses wpt directly with the new --inject-script support for injecting polyfills described at:
https://web-platform-tests.org/running-tests/testing-polyfills.html

The previous custom webdriver approach is maintained for compatibility with saucelabs.


flackr:
@egirard could you have a look as well. I've tested on a windows machine to make sure the OS independent replacements work for me but thought it couldn't hurt to have you check as well. For some reason the full wpt run doesn't work but this seems outside the scope of this - i.e. given test:simple works test:wpt *should* as long as wpt run is set up correctly.

kevers-google (approved):
LGTM

PR #118, opened 2023-07-26
Reland114

egirard:
Reland Gaubee's https://github.com/flackr/scroll-timeline/pull/114
Replaces orientation with axis to reflect updates to names in the spec.


src/index.js
```diff
@@ -23,33 +23,43 @@ import {
 
 import { initCSSPolyfill } from "./scroll-timeline-css"
 
-initCSSPolyfill();
+export function initPolyfill() {
+  if (initCSSPolyfill()) {
+    return true;
```
kevers-google, line 28: This early return skips the checks below.

src/index.js
```diff
@@ -23,33 +23,43 @@ import {
 
 import { initCSSPolyfill } from "./scroll-timeline-css"
 
-initCSSPolyfill();
+export function initPolyfill() {
```
kevers-google, line 26: Why do we need to export this if calling it directly below?

kevers-google (changes requested):

src/index.js
egirard: initCSSPolyfill returns true iff the browser natively supports SDA.  So in that case, the js polyfill (below) is also going to be skipped.  

I will add documentation to describe this.

src/index.js
egirard: Agree that this is currently unnecessary.  
There could be an argument that folks might want to know if the polyfill is active or just using the native browser implementation (so the return true at the top of the function might be interesting), but folks can't call that function safely now (since they would need to call it a second time, and it doesn't handle that gracefully.)

Pulling this out.

src/index.js
```diff
@@ -23,33 +23,44 @@ import {
 
 import { initCSSPolyfill } from "./scroll-timeline-css"
 
-initCSSPolyfill();
+function initPolyfill() {
+  // initCSSPolyfill returns true iff the host browser supports SDA
+  if (initCSSPolyfill()) {
+    return true;
```
kevers-google, line 29: Nit: Return type is ignored and will return undefined if not early exiting.  "return;" instead of "return true;".

kevers-google (changes requested):

PR #131, opened 2023-08-18
Allow a duration of `auto` in the animation shorthand

### LINKED ISSUE #128: Handle `auto` duration in the `animation` shorthand
The polyfill can’t handle this, whereas the native implementation does handle it just fine:

```css
animation: reveal auto linear both;
```

Removing `auto` from the shorthand makes the polyfill work again

```css
animation: reveal linear both;
```

bramus:
Fixes #128.


kevers-google (approved):
LGTM

PR #132, opened 2023-08-18
Add anonymous view timeline support (`view()`)

### LINKED ISSUE #129: Support anonymous View Timelines (`view()`)
The polyfill can’t handle this:

```css
animation: reveal linear both;
animation-timeline: view();
```

Switching to a named timeline works fine:

```css
view-timeline: --self;
animation: reveal linear both;
animation-timeline: --self;
```

bramus:
This PR adds support for anonymous view timelines, which fixes #129.

Marking this PR as a draft as it currently only covers `view()` and `view(axis)` support. It does not cater for `view(inset)` nor `view(axis inset)` yet.

I am using separate code paths to keep the ScrollTimeline and ViewTimeline code separated, even though both are using the same logic and structure. An alternative approach could be to merge both code paths, but then it would need some trickery in `getAnonymousTimelineOptions` to differentiate between ScrollTimelines and ViewTimelines, so that it can return the correct `options` for each type. I thought of using a different prefix for the anonymous timeline name _(using `:st` and `:vt` instead of just `:t` to differentiate between both)_ but that felt a bit hacky.

### DIFF
--- a/src/scroll-timeline-css-parser.js
+++ b/src/scroll-timeline-css-parser.js
@@ -18,7 +18,8 @@ export const RegexMatcher = {
   ANIMATION_TIME_RANGE: /animation-range\s*:([^;}]+)/,
   ANIMATION_NAME: /animation-name\s*:([^;}]+)/,
   ANIMATION: /animation\s*:([^;}]+)/,
-  ANONYMOUS_SCROLL: /scroll\(([^)]*)\)/,
+  ANONYMOUS_SCROLL_TIMELINE: /scroll\(([^)]*)\)/,
+  ANONYMOUS_VIEW_TIMELINE: /view\(([^)]*)\)/,
 };
 
 // Used for ANIMATION_TIMELINE, ANIMATION_NAME and ANIMATION regex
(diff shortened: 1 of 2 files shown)


bramus:
Figured out that adding inset support is quite easy, as the checks already happen in `scroll-timeline-base`. Marking as ready for review.

src/scroll-timeline-css-parser.js
```diff
@@ -159,10 +161,28 @@ export class StyleParser {
...
+
   getViewTimelineOptions(timelineName, target) {
+    const anonymousTimelineOptions = this.getAnonymousViewTimelineOptions(timelineName, target);
+    if(anonymousTimelineOptions)
+      return anonymousTimelineOptions;
+
     for (let i = this.subjectSelectorToViewTimeline.length - 1; i >= 0; i--) {
       const options = this.subjectSelectorToViewTimeline[i];
       if(options.name == timelineName) {
+        // TODO: The subject is the target, no? If yes, then can remove this extra check.
```
kevers-google, line 185: This block appears to be for resolving something of the form:
animation-timeline: --name;
view_timeline: --name; // In ancestor element.

In this case I believe that target is the target element for the animation, which is not necessarily the subject.



Also findPreviousSibling....Selector will need some refactoring for supporting timeline attachments as the current is using stale scoping rules.  Please add a TODO.

bramus: Pushed new commits to resolve this.

kevers-google (approved):

PR #136, opened 2023-08-25
Declare `parts` variable before using it

### LINKED ISSUE #119: Does not support strict mode?
Hello everyone, 

I made this small pen https://codepen.io/lukasgerm/pen/oNQJbWa which is using the polyfill. But the SDK is throwing an error 

```
ReferenceError: assignment to undeclared variable parts
```

It is probably related to codepen enabling strict mode by or so (just an assumption). What do you guys suggest here?

bramus:
In JavaScript Strict Mode, all variables need to be declared. Fixes #119.


kevers-google (approved):
LGTM

PR #137, opened 2023-08-25
Return the default rangeOffset when no value is given

bramus:
When no range information is passed in via the Animation’s `options`, demos such as the one in https://github.com/flackr/scroll-timeline/issues/119#issuecomment-1658247473 would throw an error: `TypeError: null is not an object (evaluating 'n.rangeName')`.

This PR makes sure that the default rangeOffset value is used when none is given.


kevers-google (approved):
LGTM

PR #138, opened 2023-08-25
Update readme

bramus:
This PR gives the README an overhaul to more correctly reflect what this polyfill is about and how/when/where to use it.


kevers-google (commented):
LGTM

PR #139, opened 2023-08-25
Add missing parseError function

bramus:
This `parseError` method is called in the CSS parser but is nowhere to be found. This PR fixes this.

The function is adapted from https://github.com/GoogleChromeLabs/container-query-polyfill/blob/faf2600f7cd7b16e500f40676dd1b144005bd186/src/engine.ts#L418, which formed the basis of the CSS Parser used in the scroll-timeline polyfill.


kevers-google (approved):
LGTM

PR #140, opened 2023-08-25
Improve WPT Score

bramus:
This PR ensures that the `axis` value gets checked when creating a new `ScrollTimeline` instance.

With this adjustment the _“Creating a ScrollTimeline with an invalid axis value should throw”_ test in `/scroll-animations/scroll-timelines/constructor.html` now passes.

### DIFF
--- a/src/scroll-timeline-base.js
+++ b/src/scroll-timeline-base.js
@@ -220,7 +220,7 @@ export class ScrollTimeline {
   constructor(options) {
     scrollTimelineOptions.set(this, {
       source: null,
-      axis: (options && options.axis) ? options.axis : 'block',
+      axis: null,
       anonymousSource: (options ? options.anonymousSource : null),
       anonymousTarget: (options ? options.anonymousTarget : null),
 
@@ -236,6 +236,10 @@ export class ScrollTimeline {
       options && options.source !== undefined ? options.source
                                               : document.scrollingElement;
     updateSource(this, source);
+
+    const axis = (options && options.axis) ? options.axis : 'block';
+    this.axis = axis;
+
     updateInternal(this);
   }
 


src/scroll-timeline-base.js
```diff
@@ -236,6 +237,12 @@ export class ScrollTimeline {
       options && options.source !== undefined ? options.source
                                               : document.scrollingElement;
     updateSource(this, source);
+
+    if ((options && options.axis !== undefined) &&
```
kevers-google, line 241: I see how this handles { axis: null } but fail to see how this handles { axis: 'nonsense' } if the ScrollTimeline is created via JavaScript.  Assuming that the invalid axis name is rejected upstream in the case of a CSS animation.  We should fix this for the programmatic case as well.  We should add explicit checking that the axis is valid before assigning here in the constructor.

let timeline = new ScrollTimeline({ axis: 'yes' });
VM202:1 Uncaught TypeError: Failed to construct 'ScrollTimeline': Failed to read the 'axis' property from 'ScrollTimelineOptions': The provided value 'yes' is not a valid enum value of type ScrollAxis.
    at <anonymous>:1:16
(anonymous) @ VM202:1

kevers-google (changes requested):

src/scroll-timeline-base.js
bramus: The check happens on the line after this by setting `this.axis` to `options.axis`. The setter for `axis` (`set axis(axis) { … }` gets invoked at that moment. It contains the necessary checking logic that throws in case an invalid value is passed into the constructor: 

https://github.com/flackr/scroll-timeline/blob/01c38f0d3a02ca5c79854400b2015a266e0a31ba/src/scroll-timeline-base.js#L251-L259

kevers-google: Ah perfect.

kevers-google (approved):

src/scroll-timeline-base.js
bramus: Turns out I was incorrect in stating that the setter gets called from the constructor. This is not the case. I’ve filed #152 as a follow-up PR to fix this.

PR #143, opened 2023-08-28
Add `entry-crossing` and `exit-crossing` ranges

bramus:
These two ranges are missing from the polyfill. This PR fixes that.

Also included is a fix to cater for scenarios where the defined `rangeStart` comes after the `rangeEnd`. This was surfaced with the view-timeline demo that uses `entry 150% exit -50%` as the range. For the taller than scrollport scenario, the `rangeEnd` comes before the `rangeStart`. I looked at what Blink does for this, and there it seems that it respects the `rangeStart` and immediately flips to the end frame when passing that offset. This behavior is no mimicked by the polyfill.

### DIFF
--- a/src/scroll-timeline-base.js
+++ b/src/scroll-timeline-base.js
@@ -506,6 +506,7 @@ export function calculateRange(phase, container, target, axis, optionsInset) {
 
   let startOffset = undefined;
   let endOffset = undefined;
+  let targetIsTallerThanContainer = viewSize > containerSize ? true : false;
 
   switch(phase) {
     case 'cover':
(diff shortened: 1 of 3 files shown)


src/scroll-timeline-base.js
```diff
@@ -506,6 +506,7 @@ export function calculateRange(phase, container, target, axis, optionsInset) {
 
   let startOffset = undefined;
   let endOffset = undefined;
+  const targetIsTallerThanContainer = viewSize > containerSize ? true : false;
```
kevers-google, line 510: Suggest we add a TODO here for handling position sticky. Naturally, we can implement position sticky in a follow up PR.

bramus: Added.

kevers-google (approved):

PR #147, opened 2023-08-29
Workaround for Safari’s BFCache not playing nice with the polyfill

### LINKED ISSUE #146: Scroll-Driven Animations in Safari hang after traversing the browser’s history … unless you have Web Inspector open
Now here’s a weird situation … in Safari, when traversing the history, animations are not restored correctly. They just hang. What’s even weirder is that from the moment you try to debug this – by opening Web Inspector - the issue is gone.

https://github.com/flackr/scroll-timeline/assets/213073/ec0adb0c-bda1-48a6-842b-41f639efbdbe

By the looks of it the animations get paused, but never tick forwards. I’m guessing there’s some caching issue going on, which needs investigation.

bramus:
This PR clear the `proxyAnimations` cache when the page gets unloaded [using the unload event](https://webkit.org/blog/516/webkit-page-cache-ii-the-unload-event/). By clearing that cache new `ProxyAnimation` instances get created when going back using the browsers’s history.

This is needed for Safari, whose BFCache doesn’t play nice with this polyfill. See https://github.com/flackr/scroll-timeline/issues/146#issuecomment-1698159183 for more details.

Fixes #146.

### DIFF
--- a/src/scroll-timeline-css.js
+++ b/src/scroll-timeline-css.js
@@ -171,4 +171,12 @@ export function initCSSPolyfill() {
       }
     });
   });
+
+  // Clear cache containing the ProxyAnimation instances when leaving the page.
+  // See https://github.com/flackr/scroll-timeline/issues/146#issuecomment-1698159183
+  // for details.
+  window.addEventListener('pagehide', (e) => {
+    console.log('pagehide');
+    proxyAnimations = new WeakMap();
+  }, false);
 }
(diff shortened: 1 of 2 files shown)


src/scroll-timeline-css.js
```diff
@@ -171,4 +171,12 @@ export function initCSSPolyfill() {
       }
     });
   });
+
+  // Clear cache containing the ProxyAnimation instances when leaving the page.
+  // See https://github.com/flackr/scroll-timeline/issues/146#issuecomment-1698159183
+  // for details.
+  window.addEventListener('pagehide', (e) => {
+    console.log('pagehide');
```
kevers-google, line 179: Remove console.log.  Assume was put in for debugging.

kevers-google (changes requested):

kevers-google (approved):

PR #152, opened 2023-09-13
Validate `axis` in constructor + fix axis update regression

bramus:
Merging #140 introduced a regression making https://flackr.github.io/scroll-timeline/demo/view-timeline/ no longer work. Upon updating the axis, it always remained `block` (the default value). This is because the `this.axis` was directly manipulated, instead of storing it in the options.

Furthermore, the setter – which validates the axis – didn’t get called in the constructor, so we need to explicitly validate the axis in the constructor as well.

This PR fixes all that.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -12,3 +12,7 @@ export function parseLength(obj, acceptStr) {
   }
   return null;
 }
+
+export function isValidAxis(axis) {
+  return ["block", "inline", "x", "y"].includes(axis);
+}
--- a/src/utils.js
+++ b/src/utils.js
@@ -12,7 +12,3 @@ export function parseLength(obj, acceptStr) {
   }
   return null;
 }
-
-export function isValidAxis(axis) {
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -12,3 +12,7 @@ export function parseLength(obj, acceptStr) {
   }
   return null;
 }
+
+export function isValidAxis(axis) {
```
kevers-google, line 16: This could be put into the ScrollTimeline class as you can call this.method inside the constructor.  If we want to make it a static, then we can add it as a method to ScrollTimeline's prototype. This approach might be even better.

```
ScrollTimeline.isValidAxis(axis);
```

This results in less exports to deal with and provides better encapsulation.

kevers-google (changes requested):

src/utils.js
bramus: .done

kevers-google (approved):
LGTM

PR #153, opened 2023-09-13
Add `animation-range` support for ScrollTimeline

### LINKED ISSUE #151: Is animation-range for scroll-timeline not supported?
I have tried using both JS and CSS methods but cannot get `rangeStart:`, `rangeEnd` or `animation-range` to work in Safari or Firefox (where this polyfill would be needed).

For a reproduction, I cloned the repo, ran `npm i`, then edited `demo/basic/anonymous-scroll-timeline.html` to add:

```css
...
 #box_one {
    width: 100px;
    height: 100px;
    background-color: green;
    animation: move linear;
    animation-timeline: scroll(y nearest);
    animation-range: 40px 150px;  /* <--- ADDED THIS LINE */
  }
...
```

In Chrome, this had the intended effect (att first, scrolling does not animate anything, then the box suddenly moves swiftly downward and right). In Safari and Firefox the behaviour is the same as before adding that line. 

If the lack of support for animation-range is intentional, could you suggest a workaround? (For context, what I'm hoping to achieve is a fixed-position large, heroic header, which shrinks down to a small top-bar as one scrolls further down the page.)

bramus:
This PR adds support for `animation-range` for ScrollTimeline based SDAs. It is done by parsing the animation range _(with a fallback to the defaults of 0% and 100%)_ and then calculating the relative ScrollTimeline position. This flow has been copied over from how ViewTimelines are being handled.

It also includes a little refactoring that splits off `normalizeAxis` into its own function, as the logic to determine that ended up being present 3 times in the source code.

Fixes #151.


src/proxy-animation.js
```diff
@@ -2,8 +2,10 @@ import {
   ScrollTimeline,
   addAnimation,
   removeAnimation,
-  relativePosition
+  relativeViewTimelinePosition,
```
kevers-google, line 5: How about making relativePosition a member of ScrollTimeline with an override in ViewTimeline.  This approach would provide better encapsulation. The relativePosition method was added before non-view ScrollTimelines could have an animation range.  Then we can simply call this.timeline.relativePosition(offset) rather than handling view and scroll timelines separately.

kevers-google (dismissed):

src/proxy-animation.js
bramus: .done

bramus:
Addressed the remarks (+ merge conflict), but am wondering: maybe I should also move the `parseAnimationRange` functions onto the classes as statics?

ryantownsend:
+1 for this issue – great to see the PR @bramus 

My usage includes viewport units which I can see aren't included in the PR, would that be straight-forward to introduce? I'm not asking you to do it, I'm happy to contribute if I can, I'm just unfamiliar with the codebase – I just figure the responsive nature of viewport units could be tricky if the values are evaluated on discovery rather than in 'real time'.

bramus:
@ryantownsend I once started a PR to add viewport units – https://github.com/flackr/scroll-timeline/pull/52 – but that one has stalled. Haven’t found time to further look into it, but feel free to pick it up. It’s quite an old PR, so does need a rebase …

kevers-google:
Hi Ryan,
If you would like to contribute, I'll be happy to review the code.  Alas, I'm a bit short on free cycles, or I'd dive in and help tidy things up.

ryantownsend:
Thanks @bramus / @kevers-google – I'll see if I get some time over the next week or so.

bramus:
Converting to draft as I noticed a bug in the calculations … need to dig into that.

bramus:
FYI: I’m awaiting some of the refactors from @johannesodland’s PRs before picking this one up again.

bramus:
I think I re-landed the changes of the original branch on top of `master`. As I suspected, the issue with the wrong range being applied got fixed thanks to #176 _(confirmed through manual testing)_

The tests, however, don’t seem to have made a huge jump: only 2 extra tests pass. Also seeing timeouts for the tests with the timeline-range … that needs to be looked into.

test/expected.txt
```diff
@@ -798,14 +798,14 @@ FAIL	/scroll-animations/scroll-timelines/scroll-animation.html	Sending animation
...
-FAIL	/scroll-animations/scroll-timelines/scroll-timeline-range.html	Scroll timeline with percentage range [JavaScript API]
+PASS	/scroll-animations/scroll-timelines/scroll-timeline-range.html	Scroll timeline with percentage range [JavaScript API]
 FAIL	/scroll-animations/scroll-timelines/scroll-timeline-range.html	Scroll timeline with px range [JavaScript API]
 FAIL	/scroll-animations/scroll-timelines/scroll-timeline-range.html	Scroll timeline with calculated range [JavaScript API]
 FAIL	/scroll-animations/scroll-timelines/scroll-timeline-range.html	Scroll timeline with EM range [JavaScript API]
-FAIL	/scroll-animations/scroll-timelines/scroll-timeline-range.html	Scroll timeline with percentage range [CSS]
-FAIL	/scroll-animations/scroll-timelines/scroll-timeline-range.html	Scroll timeline with px range [CSS]
-FAIL	/scroll-animations/scroll-timelines/scroll-timeline-range.html	Scroll timeline with calculated range [CSS]
-FAIL	/scroll-animations/scroll-timelines/scroll-timeline-range.html	Scroll timeline with EM range [CSS]
+TIMEOUT	/scroll-animations/scroll-timelines/scroll-timeline-range.html	Scroll timeline with percentage range [CSS]
```
johannesodland, line 805: I tried to look into why these tests time out. It seems like a combination of lenient parsing, and reuse of the source elements in the tests. 

We parse and store values for range start/end that we cannot resolve yet, such as `5em`.
We then fail to auto-align the start time, as the offsets can't be resolved, and throw an error.

We could probably both
- be stricter in values we accept and throw a TypeError straight away when setting the value
- handle runtime errors in stead leaving the animation in an unexpected state


This is not something we need to handle now, but I can open a separate PR for handling unresolvable ranges when auto-aligning the start time.

src/scroll-timeline-base.js
```diff
@@ -863,4 +906,112 @@ export class ViewTimeline extends ScrollTimeline {
     return CSS.px(range(this,'cover').end);
   }
 
+  // Calculate the fractional offset of a (phase, percent) pair relative to the
+  // full cover range.
+  relativePosition(value, details) {
+    const { rangeName, offset } = value;
+    
+    // @TODO: Precalc and store these
```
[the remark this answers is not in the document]
bramus: This gets called a lot, so I’m quite sure that caching these will have its benefits.

src/scroll-timeline-base.js
```diff
@@ -476,6 +477,62 @@ export class ScrollTimeline {
   static isValidAxis(axis) {
     return ["block", "inline", "x", "y"].includes(axis);
   }
+
+  // Calculate the fractional offset of a range value relative to the full range.
+  relativePosition(value, details) {
+    const { axis, source } = details.timeline;
+
+    // @TODO: Make use of sourceMeasurements here, yet these don’t seem to be stored
```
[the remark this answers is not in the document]
bramus: Thanks! Updating …

src/scroll-timeline-base.js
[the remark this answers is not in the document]
flackr: As long as it's not calling out to any DOM apis, i.e. using all cached geometry, a pure JS function is quite fast. I.e. unless this is being called 100s of times per frame it's unlikely to have a substantial impact. However, if you find that caching makes a noticeable difference on some pages I'm open to it.

package.json
```diff
@@ -8,9 +8,9 @@
     "build": "microbundle -f iife",
     "dev": "run-all \"serve\" \" microbundle watch -f iife \"",
     "deploy": "npm run build",
-    "test-setup": "node test/setup/checkout-wpt.mjs",
     "test:wpt": "npm run test-setup && cd test && cd wpt && (python wpt run --headless -y --log-wptreport ../report/data.json --log-wptscreenshot=../report/screenshots.txt --log-html=../report/index.html --inject-script ../../dist/scroll-timeline.js firefox scroll-animations || true)",
     "test:simple": "npm run test-setup && cd test && cd wpt && python wpt serve --inject-script ../../dist/scroll-timeline.js",
+    "test-setup": "node test/setup/checkout-wpt.mjs && npm run build",
```
flackr, line 13: I'm surprised you regularly use test-setup. You want to pull the latest from wpt as part of your build and test flow?

If we make `npm run build` part of the `test-setup` can you remove the extra call to `npm run build` from gh-pages.yml?

src/scroll-timeline-base.js
```diff
@@ -476,6 +477,61 @@ export class ScrollTimeline {
   static isValidAxis(axis) {
     return ["block", "inline", "x", "y"].includes(axis);
   }
+
+  // Calculate the fractional offset of a range value relative to the full range.
+  relativePosition(value, details) {
```
flackr, line 482: It looks like there are a few previous existing offenders that slipped in, but the goal of the polyfill is to present only the exact interface of ScrollTimeline / ViewTimeline. Any methods / helpers which aren't part of the public javascript interface can be free functions that accept the timeline as a parameter, e.g. see the [updateInternal function](https://github.com/flackr/scroll-timeline/blob/147d187620e49ce5fb3abc9e1c3e1018755d8f02/src/scroll-timeline-base.js#L34).

src/scroll-timeline-base.js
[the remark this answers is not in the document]
bramus: I’ll leave it at this for now.

package.json
bramus: I don’t directly run `npm run test-setup`, but do want a fresh build when running `test:wpt` or `test:simple` which rely on `test-setup`. That’s why I added it here.

src/scroll-timeline-base.js
bramus: Got it.

package.json
flackr: Ah, makes sense.

src/proxy-animation.js
```diff
@@ -960,8 +950,7 @@ export class ProxyAnimation {
       effect: null,
       // Range when using a view-timeline. The default range is cover 0% to
```
flackr, line 951: nit: Update the comment.

bramus: Done.

src/scroll-timeline-base.js
bramus: Done.

package.json
bramus: Done.

bramus:
Addressed all remarks. Requesting a new review, @flackr @johannesodland

src/proxy-animation.js
```diff
@@ -912,6 +906,147 @@ function autoAlignStartTime(details) {
...
+
+  if (timeline instanceof ViewTimeline) {
+    // Format:
+    // <start-name> <start-offset> <end-name> <end-offset>
+    // <name> --> <name> 0% <name> 100%
+    // <name> <start-offset> <end-offset> --> <name> <start-offset>
+    //                                        <name> <end-offset>
+    // <start-offset> <end-offset> --> cover <start-offset> cover <end-offset>
+    // TODO: Support all formatting options once ratified in the spec.
+    const parts = value.split(' ');
```
johannesodland, line 954: https://github.com/flackr/scroll-timeline/pull/209 adds a function `splitIntoComponentValues()` to split a string of component values into separate strings taking functions and parenthesis into account. 

If https://github.com/flackr/scroll-timeline/pull/209 is merged first, we could use this function here, avoiding issues like `cover calc(0% - 10px) cover calc(100% + 10px)` being split into `['cover', 'calc(0%', '-', '10px)', 'cover', 'calc(100%', '+', '10px)']`

But, we could always do this at a later point.

src/proxy-animation.js
```diff
@@ -912,6 +906,147 @@ function autoAlignStartTime(details) {
...
+    const rangeNames = [];
+    const offsets = [];
+  
+    parts.forEach(part => {
+      if (part.endsWith('%'))
```
johannesodland, line 959: To support `calc()`, maybe this could be replaced by 

```js
if (ANIMATION_RANGE_NAMES.includes(part)) 
  rangeNames.push(part);
} else {
  offsets.push(CSSNumericValue.parse(part));
}
```

Then the following code can be removed: 

```js
    // TODO: allow calc() in the offsets
     if (offsets.length > 1) {
       animationRange.start.offset = CSS.percent(offsets[0]);
       animationRange.end.offset = CSS.percent(offsets[1]);
     }
```

src/proxy-animation.js
```diff
@@ -912,6 +906,147 @@ function autoAlignStartTime(details) {
...
+      animationRange.end.offset = CSS.percent(offsets[1]);
+    }
+  
+    return animationRange;
+  }
+
+  if (timeline instanceof ScrollTimeline) {
+    // @TODO: Play nice with only 1 offset being set
+    // @TODO: Play nice with expressions such as `calc(50% + 10px) 100%`
+    const parts = value.split(' ');
```
johannesodland, line 986: Same as above: 

https://github.com/flackr/scroll-timeline/pull/209 adds a function splitIntoComponentValues() to split a string of component values into separate strings taking functions and parenthesis into account.

src/proxy-animation.js
```diff
@@ -912,6 +906,147 @@ function autoAlignStartTime(details) {
...
+        rangeName = value.rangeName;
+      }
+
+      if (value.offset !== undefined) {
+        offset = value.offset;
+      }
+    }
+    // Author passed in something like `"cover 100%"`
+    else {
+      const parts = value.split(rangeNameRegExp).map(part => part.trim()).filter(Boolean);
```
johannesodland, line 1020: Sorry for bringing up https://github.com/flackr/scroll-timeline/pull/209 a lot 😅 

If it's merged, we could clean this up:
```js
const parts = splitIntoComponentValues(value);
```

src/scroll-timeline-base.js
```diff
@@ -780,16 +777,36 @@ function calculateInset(value, sizes) {
   return { start, end };
 }
 
+// Calculate the fractional offset of a range value relative to the normal range.
+export function fractionalOffset(timeline, value) {
```
bramus, line 781: FYI: I renamed this function from `relativePosition` to `fractionalOffset` for two reasons:

1. More clearly expresses what it does
2. There’s already a `relativePosition` function in another file present.

johannesodland (approved):
LGTM 👍 

I've added some comments with regards to https://github.com/flackr/scroll-timeline/pull/209 that adds a function to split a string into component values. 

If we merge #209 first we should use that function here, otherwise I can include those changes in #209.

package.json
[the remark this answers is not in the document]
bramus: I think it’s best if I remove these build step changes here, and let them be part of a bigger PR that revisits the npm scripts structure (and build pipeline in general) – I feel there’s a bigger discussion to be done here.

src/proxy-animation.js
bramus: Depending on which one gets in first, I’ll ask for your help to fix these here.

_(For my convenience it would be easiest if this one makes it in first 😝)_

johannesodland: > (For my convenience it would be easiest if this one makes it in first 😝)

Let's do it that way 😀👍

flackr, line 954: It'd be a bit better to split on \s+ instead of a space so that any number of whitespace characters are skipped.

flackr (approved):
Looks good.

src/proxy-animation.js
bramus: That should to be fixed in https://github.com/flackr/scroll-timeline/pull/209 by @johannesodland.

bramus: Fixed. Tests pass locally. Will merge once tests here confirm.

Thanks for the review!

PR #154, opened 2023-09-13
Stop processing ScrollTimeline props if no animation-related properties have been declared

bramus:
While debugging something for #153, I noticed that `saveInRelationList` was being called once per ruleset – even though only 1 of the processed ruleset in my test had animation-related properties declared.

This PR adds a quick perf win by bailing out early from the `handleScrollTimelineProps` method in case no animation-related properties have been extracted.


kevers-google (changes requested):
Format nit.  Let's keep to Google style guidelines.
```
if (...) {
  return;
}
```

space after if and statement on a separate line with {} even though 1 line.  I'm sure, that I've violated this elsewhere, but no sense proliferating inconsistencies.

kevers-google (approved):

Issue #156, opened 2023-09-17
prefers-reduced-motion `matches` throws on animationstart handler on older Chrome versions


calinoracation:
**Chrome Version:** 106 (Cypress unit tests caught this)
**Full Error:**
```
Uncaught DOMException: Failed to execute 'matches' on 'Element': '@media (prefers-reduced-motion:reduce)' is not a valid selector.
```

**Source of issue in scroll-timeline-css-parser:**
```
        if (target.matches(current.selector)) { // <-- This line throws
          if (!current['animation-name'] || current['animation-name'] == animationName) {
            return {
              'animation-timeline': current['animation-timeline'],
              'animation-range': current['animation-range'],
            };
          }
        }
```
I can open a PR with wrapping a try/catch around it and let the iteration continue without throwing, but wanted to report the issue directly first.

bramus:
This isssue is caused by the CSS parser that is used not being able to handle everything. While a different at-rule here, it is exactly the same as `@layer` not being parseable: https://github.com/flackr/scroll-timeline/issues/80

Issue #80, opened 2022-09-23
Polyfill does not play nice with @layer


bramus:
When using some of the props inside `@layer` …

```css
@layer foo {
	@supports (animation-timeline: scroll()) {
		/* Hide Warning */
		.warning {
			display: none;
		}
	}
}
```

… it throws this:

```
Uncaught DOMException: Failed to execute 'matches' on 'Element': '@layer foo' is not a valid selector.
    at t.getAnimationTimelineOptions (https://flackr.github.io/scroll-timeline/dist/scroll-timeline.js:1:25932)
    at https://flackr.github.io/scroll-timeline/dist/scroll-timeline.js:1:32971
    at https://flackr.github.io/scroll-timeline/dist/scroll-timeline.js:1:33181
    at Array.forEach (<anonymous>)
    at https://flackr.github.io/scroll-timeline/dist/scroll-timeline.js:1:32922
```

When not using a `@layer`, it works fine.

```css
@supports (animation-timeline: scroll()) {
	/* Hide Warning */
	.warning {
		display: none;
	}
}
```

bramus:
It’s not related to the `@supports` rule BTW. This also throws:

```css
@layer foo {
	.revealing-image {
		view-timeline-name: revealing-image;
		view-timeline-axis: block;

		animation: linear reveal both;
		animation-timeline: revealing-image;

		/* Needed for polyfill (wrong syntax): */
		animation-time-range: enter 25% 100%;

		/* Needed for Chrome Canary: */
		animation-range: enter 25% 100%;
		animation-delay-start: enter 25%;
		animation-delay-end: enter 100%;
	}
}
```

flackr:
I think `@layer` could be safely ignored for the mostpart - as long as we are picking up the inner rules. Though strictly speaking we should reorder the rules according to the defined layer order.

For `@media` and we should probably check whether the media query matches (e.g. `window.matchMedia`), and for `@supports` we could use `CSS.supports`.

flackr:
I wrote a CSS parser at https://github.com/flackr/carousel/blob/main/polyfill/polyfill.js which can correctly parse `@media`, `@layer` and nested selectors. I think I'll bring that over to this project as it should be more robust.

As for test coverage, I'm not sure if there's much (or any) wpt tests of scroll timelines using nested selectors as it's normally expected that features can be tested independently. Maybe this means we need a scroll-timeline polyfill specific test suite for things that don't make sense to be in wpt.

PR #157, opened 2023-09-18
Fix issue with target.matches in CSS parsing

calinoracation:
Fixes https://github.com/flackr/scroll-timeline/issues/156. Adds a try/catch and allows to continue matching if a selector throws.


bramus:
While this does make the error go away, it does not fix the root cause, which is the CSS parser not being able to parse nested at-rules. Also see #80.

Could be merged for now, but in that case I’d change the `// Continue matching` comment to `// TODO: handle nested at-rules`

calinoracation:
@bramus Ah that makes sense, I updated the TODO.

Issue #165, opened 2023-10-19
Polyfill does not account for subject borders


johannesodland:
The polyfill uses `clientWidth/height` in when calculating the size of the subjects principal box, but `clientWidth/height` does not include borders. This results in errors when calculating ranges for ViewTimeline.

Result with polyfill:
<img width="536" alt="image" src="https://github.com/flackr/scroll-timeline/assets/175195/b1f026db-f5b8-460b-a280-15a8dbe2c651">

Result with Chrome:
<img width="549" alt="image" src="https://github.com/flackr/scroll-timeline/assets/175195/6a195d7e-8925-4e6f-9693-60a47bc4f093">

PR #166, opened 2023-10-19
Account for subject borders

johannesodland:
This PR fixes https://github.com/flackr/scroll-timeline/issues/165

The polyfill used clientWidth/Height to calculate the size of the principal box of the subject, but clientWidth/Height does not include borders. This caused errors when calculating ranges.

### DIFF
--- a/src/scroll-timeline-base.js
+++ b/src/scroll-timeline-base.js
@@ -481,14 +481,14 @@ export function calculateRange(phase, container, target, axis, optionsInset) {
   if (axis == 'x' ||
       (axis == 'inline' && horizontalWritingMode) ||
       (axis == 'block' && !horizontalWritingMode)) {
-    viewSize = target.clientWidth;
+    viewSize = target.offsetWidth;
     viewPos = left;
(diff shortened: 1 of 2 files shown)


src/scroll-timeline-base.js
```diff
@@ -473,6 +473,7 @@ export function calculateRange(phase, container, target, axis, optionsInset) {
   // The view position is the scroll position of the logical starting edge
   // of the view.
   const style = getComputedStyle(container);
+  const targetBCR = target.getBoundingClientRect();
```
flackr, line 476: getBoundingClientRect is transformed through all of the ancestor transforms applying to that element. This won't give the correct result if there are any scale transforms applied. You should use offsetWidth / offsetHeight instead of clientWidth / clientHeight as these return what we want, the element's local box size including borders according to https://drafts.csswg.org/cssom-view/#dom-htmlelement-offsetwidth

johannesodland: Thanks 🙏 
I've updated the PR.

Issue #169, opened 2023-11-03
CSSMathNegate and CSSMathInvert should have a read only value attribute


johannesodland:
All CSSMathValues in the proxy-cssom have a readonly `values` attribute returning a CSSNumericArray, but [CSSMathNegate](https://www.w3.org/TR/css-typed-om-1/#cssmathnegate) and [CSSMathInvert](https://www.w3.org/TR/css-typed-om-1/#cssmathinvert) should have a singular `value` attribute [according to the spec](https://www.w3.org/TR/css-typed-om-1/#complex-numeric).

Code attempting to read the value attribute in browsers where proxy-cssom is used will fail.

This is not a full CSS Typed OM polyfill, but if we want to support using calculations in ranges, we need to be able to read the value attribute.

bramus:
With #170 being merged, this issue may be closed now.

PR #170, opened 2023-11-03
Add value getter to CSSMathInvert and CSSMathNegate

johannesodland:
Solves #169

This PR adds a getter for `value` to the CSSMathInvert and CSSMathNegate classes. 

While this does not remove the `values` getter on the MathOperation superclass, it makes sure that code calling .value on instances of these classes won't fail.


kevers-google (approved):
Looks good

PR #171, opened 2023-11-05
Support CSSMathValue as rangeStart and rangeEnd

johannesodland:
Adding support for CSSMathValues in rangeStart and rangeEnd if they can be simplified to a CSSUnitValue with a `px` unit. 

The following WPT subtests should now pass: 
- View timeline with range as <name> <px> pair.
- View timeline with range as <name> <percent+px> pair.

This is solved by partially implementing [`simplify a calculation tree`](https://www.w3.org/TR/css-values-4/#simplify-a-calculation-tree) and partially implement support for a few more methods ([`type()`](https://drafts.css-houdini.org/css-typed-om/#dom-cssnumericvalue-type), [`to()`](https://drafts.css-houdini.org/css-typed-om/#dom-cssnumericvalue-to) and  [`toSum()`](https://drafts.css-houdini.org/css-typed-om/#dom-cssnumericvalue-tosum)) in the CSSOM/Typed OM proxy.

Note: 
This implementation limited to using CSS Typed OM for rangeStart and rangeEnd, and does not support parsing of math functions. (This is something we can look into at a later stage. )

The implementation does not include support for relative length units, as we would need a mechanism for renormalizing timing when the lengths they are relative to changes. It might be better to solve renormalizing timing when source and target has been resized first.

Happy to receive any feedback :)

Before PR:
image

After PR:
image

### DIFF
--- a/src/simplify-calculation.js
+++ b/src/simplify-calculation.js
@@ -1,0 +1,301 @@
+import {isCanonical} from "./utils";
+
+/**
+ * @typedef {{percentageReference: CSSUnitValue}} Info
+ */
+
+/**
+ * Groups a list of objects by a given string keyed property
+ *
+ * @template T
+ * @param {T[]} items
+ * @param {string} key string key
+ * @return {Map<any,T[]>}
+ */
+function groupBy(items, key) {
+  return items.reduce((groups, item) => {
+    if (groups.has(item[key])) {
+      groups.get(item[key]).push(item);
+    } else {
+      groups.set(item[key], [item]);
+    }
+    return groups;
+  }, new Map());
+}
+
+/**
+ * Partitions a list into a tuple of lists.
+ * The first item in the tuple contains a list of items that pass the test provided by the callback function.
+ * The second item in the tuple contains the remaining items
+ *
+ * @template T
+ * @param {T[]} items
+ * @param {(item:T) => boolean} callbackFn Returns truthy if item should be put in the first list in the tuple, falsy if it should be put in the second list.
+ * @return {[T[],T[]]}
+ */
+function partition(items, callbackFn) {
+  const partA = [];
+  const partB = [];
+  for (const item of items) {
+    if (callbackFn(item)) {
+      partA.push(item);
+    } else {
+      partB.push(item);
+    }
+  }
+  return [partA, partB];
+}
+
+/**
+ * Partial implementation of `simplify a calculation tree` applied to CSSNumericValue
+ * https://www.w3.org/TR/css-values-4/#simplify-a-calculation-tree
+ *
+ * @param {CSSNumericValue} root
+ * @param {Info} info information used to resolve
+ * @return {CSSNumericValue}
+ */
+export function simplifyCalculation(root, info) {
+  function simplifyNumericArray(values) {
+    return Array.from(values).map((value) => simplifyCalculation(value, info));
+  }
+
+  // To simplify a calculation tree root:
+  if (root instanceof CSSUnitValue) {
+    // 1. If root is a numeric value:
+
+    if (root.unit === "percent" && info.percentageReference) {
+      // 1. If root is a percentage that will be resolved against another value, and there is enough information
+      //    available to resolve it, do so, and express the resulting numeric value in the appropriate canonical unit.
+      //    Return the value.
+      const resolvedValue = (root.value / 100) * info.percentageReference.value;
+      const resolvedUnit = info.percentageReference.unit;
+      return new CSSUnitValue(resolvedValue, resolvedUnit);
+    }
+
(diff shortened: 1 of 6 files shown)


src/numeric-value-typing.js
```diff
@@ -0,0 +1,282 @@
...
+  for (const unit of group.units) {
+    unitToCompatibleUnitsMap.set(unit, group);
+  }
+}
+
+export function getSetOfCompatibleUnits(unit) {
+  return unitToCompatibleUnitsMap.get(unit);
+}
+
+function productOfUnionMaps(units1, units2) {
```
kevers-google, line 42: Comments within the method are helpful.  Can you also provide a comment above the function declaration providing a high level description of the purpose of the function perhaps with an example of sample input and expected output.

src/numeric-value-typing.js
```diff
@@ -0,0 +1,282 @@
...
+    } else {
+      // 2. Otherwise, set result[unit] to power.
+      result[unit] = units2[unit];
+    }
+  }
+  // 3. Return result.
+  return result;
+}
+
+export function createSumValue(self) {
```
kevers-google, line 59: The parameter name 'self' seems a bit unusual, since it is normally used to indicate a context.  How about calling the parameter cssNumericValue.

src/numeric-value-typing.js
kevers-google: Comment applies to the following methods as well. Though most of the others seem to be pretty clear, a function level description (perhaps with a spec link) would nonetheless be helpful.

src/simplify-calculation.js
```diff
@@ -0,0 +1,270 @@
+import {isCanonical} from "./utils";
+
+function groupBy(items, key) {
```
kevers-google, line 3: Please add function-level descriptions.

Is this all original work or are there sources that should be cited here?

src/simplify-calculation.js
```diff
@@ -0,0 +1,270 @@
...
+    if (callbackFn(item)) {
+      partA.push(item);
+    } else {
+      partB.push(item);
+    }
+  }
+  return [partA, partB];
+}
+
+export function simplifyCalculation(root, info) {
```
kevers-google, line 24: Spec-link would be helpful.

src/utils.js
```diff
@@ -11,4 +11,10 @@ export function parseLength(obj, acceptStr) {
     return new CSSUnitValue(value, unit);
   }
   return null;
+}
+
+const canonicalUnits = new Set(["px", "deg", "s", "hz", "dppx", "number", "fr"]);
```
kevers-google, line 16: Wondering if we should group these sets of units in one place.  I see canonical units here,  units by type in proxy-cssom and compatible unit types in numeric-value-typing. Not strictly required, but might be easier to spot omissions if in one place.

src/numeric-value-typing.js
```diff
@@ -0,0 +1,282 @@
+const failure = null;
```
kevers-google, line 1: Do we need to credit/acknowledge any external sources?

src/numeric-value-typing.js
johannesodland: Added a function level description with a link to the spec for each.

src/numeric-value-typing.js
johannesodland: Sounds good. Renamed the parameter 👍

src/simplify-calculation.js
johannesodland: Fixed, and rewrote to use a map so that it can handle properties with non-string values. 

I wrote the utility-function, if that's what you're asking?

src/simplify-calculation.js
johannesodland: Agree. Fixed.

src/utils.js
johannesodland: Thanks for pointing this out🙏

I moved most of the units into a common const: https://github.com/flackr/scroll-timeline/blob/ce5f19d9f0c0e012ae53d22fbb3290690aebebc0/src/numeric-values.js#L12

I renamed the file to numeric-values, as it contained both common numeric operations and numeric value typings.

I've tried to keep code for polyfilling the typed OM separate from the rest, as this will only be needed in some browsers. As a consequence there's a small duplication: I have kept the list of canonical units for lookup during simplification.

src/numeric-value-typing.js
johannesodland: I think all of the standards are now referenced in comments above the functions. Outside of that, I don't think there's other external sources? Most of this PR is a partial implementation of `simplify a calculation tree` and the bits of css-typed-om necessary to perform that simplification.

johannesodland:
Are there a preference for single or double quotes for string literals? 
I'm happy to adjust.

kevers-google (approved):

PR #172, opened 2023-11-06
Renormalize timing on source resize

johannesodland:
When the source element is resized, the specified timing is outdated and needs to be renormalized. 

This PR solves this by using a ResizeObserver to observe the source element, and trigger a renormalization whenever the size has changed.


kevers-google (approved):
Looks great.

PR #174, opened 2023-11-13
Defer tickAnimation to requestAnimationFrame to prevent relayout and ResizeObserver errors

johannesodland:
In updateInternal we call .currentTime before updating animations. 

When there's multiple ViewTimelines, updateInternal will be called individually for each timeline on the scroll event, causing unnecessary reflow. It also causes ResizeObserver to throw a "ResizeObserver loop completed with undelivered notifications." error.

Deferring tickAnimation to requestAnimationFrame to complete all calls to currentTime before ticking animations.

Example of unexpected errors in /scroll-timelines/finish-animation.html:

**Before**
image

**After**
image

### DIFF
--- a/src/scroll-timeline-base.js
+++ b/src/scroll-timeline-base.js
@@ -34,9 +34,11 @@ function updateInternal(scrollTimelineInstance) {
   let animations = details.animations;
   if (animations.length === 0) return;
   let timelineTime = scrollTimelineInstance.currentTime;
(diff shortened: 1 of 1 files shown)


johannesodland:
I realized that this will not work if we have consumers that expect timing to be updated on the next animation frame. 
Closing this PR. Will be submitting a new where we handle scroll and resize events only once for each scroll source, so that we can update the timing of all associated timelines before ticking their animations.

Issue #183, opened 2023-12-23
Broken recently


marcfilleul:
Hi,

I just checked websites where it was working correctly before and saw it wasn't working anymore.

Here are the console issues on Safari:

![CleanShot 2023-12-23 at 14 00 35@2x](https://github.com/flackr/scroll-timeline/assets/3406688/90a4b620-7b7f-423f-aaf0-da3256bbfc16)

And Firefox:

![CleanShot 2023-12-23 at 14 04 14@2x](https://github.com/flackr/scroll-timeline/assets/3406688/347b2edf-e842-4fbe-8d31-80e25628a5a3)

Looks like it could be cause by recent updates around proxy-animations.

Best regards

bramus:
Thanks for reporting. Things seems to have gone wrong indeed …

The first error about `proxyAnimations` is an error that had gone unnoticed as it only triggers in strict mode. This happens when you load the script as via `type="module"`. I’ve got PR #184 up to fix this.

The second error is a bug introduced in #178, which broke the parsing of `animation-range` for ViewTimeline. I guess this went unnoticed as none of the demos actually uses an `animation-range`. I’ve got PR #185 up to fix that bug.

/cc @johannesodland @flackr

bramus:
@marcfilleul The PRs I mentioned got merged. I believe these resolve the issues you reported. Can you confirm?

flackr:
Thanks so much for the fixes @bramus

> The first error about `proxyAnimations` is an error that had gone unnoticed as it only triggers in strict mode. This happens when you load the script as via `type="module"`. I’ve got PR #184 up to fix this.

I filed #188 to ensure we test running the code in strict mode so this sort of thing should be caught.

> The second error is a bug introduced in #178, which broke the parsing of `animation-range` for ViewTimeline. I guess this went unnoticed as none of the demos actually uses an `animation-range`. I’ve got PR #185 up to fix that bug.

We're running all of the scroll-animation wpt tests too, I wonder if it's one that we fail for some other reason already which resulted in it not being caught.

johannesodland:
> We're running all of the scroll-animation wpt tests too, I wonder if it's one that we fail for some other reason already which resulted in it not being caught.

I’ve got some time over the weekend. I’ll try to figure out which tests should have covered it, and see if we can get them to pass.

marcfilleul:
> @marcfilleul The PRs I mentioned got merged. I believe these resolve the issues you reported. Can you confirm?

[...]

PR #184, opened 2023-12-23
Proxied Animations Cleanup fixup

bramus:
In #89 the proxied animations handling got moved around a bit. In #147 (which was merged earlier) a cleanup mechanism for the old `proxyAnimations` was introduced, but this mechanism was not removed in #89. This went unnoticed as it only triggers an error when running in strict mode, as reported in #183.

This PR fixes that.

### DIFF
--- a/src/scroll-timeline-css.js
+++ b/src/scroll-timeline-css.js
@@ -159,11 +159,4 @@ export function initCSSPolyfill() {
       }
     });
   });
-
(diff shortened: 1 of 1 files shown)


flackr:
Thanks for the fix! I'll look into ensuring that the proxy is tested in strict mode so that this would have been caught by the tests.

PR #191, opened 2023-12-28
Handle undefined options in updateTimingHandler

johannesodland:
Handle case where `effect.updateTiming()` is called with an undefined input.

This happens in the [view-timeline-subject-size-changes test](https://github.com/web-platform-tests/wpt/blob/a95a859194890421952a3105e9b69789dbb4c58b/scroll-animations/view-timelines/view-timeline-subject-size-changes.html#L53).

The PR does not fix the entire test, but it fixes the following error: 
```
promise_test: Unhandled rejection with value: object "TypeError: options is undefined"
```

I will create a separate PR for fixing the test later on.

### DIFF
--- a/src/proxy-animation.js
+++ b/src/proxy-animation.js
@@ -779,6 +779,9 @@ function createProxyEffect(details) {
   };
   const updateTimingHandler = {
     apply: function(target, thisArg, argumentsList) {
+      if (!argumentsList || !argumentsList.length)
+        return;
+
       // Additional validation that is specific to scroll timelines.
       if (details.timeline) {
         const options = argumentsList[0];


src/proxy-animation.js
```diff
@@ -780,7 +780,7 @@ function createProxyEffect(details) {
   const updateTimingHandler = {
     apply: function(target, thisArg, argumentsList) {
       // Additional validation that is specific to scroll timelines.
-      if (details.timeline) {
+      if (details.timeline && argumentsList[0]) {
```
bramus, line 783: I see that further down you have this check:

```js
if (typeof duration !== 'undefined' && duration !== 'auto') {
```

This check for `undefined` can be omitted then I guess?

---

Furthermore, wouldn’t it be possible to do this check for `argumentsList[0]` earlier, and have it exit prematurely if `argumentsList.length` equals `0`?

I.e.
```js
apply: function(target, thisArg, argumentsList) {
  if (!argumentsList || !argumentsList.length) return;

  if (details.timeline) {
  …
```

johannesodland: > I see that further down you have this check:
> 
> ```js
> if (typeof duration !== 'undefined' && duration !== 'auto') {
> ```
> 
> This check for `undefined` can be omitted then I guess?
> 

I don't think we can omit that check. 

`updateTiming()` is used to perform a delta update to the timing, and the properties of the [OptionalEffectTiming](https://drafts.csswg.org/web-animations-1/#dictdef-optionaleffecttiming) argument don't have to be set.

We should only remove the autoDurationEffect flag if the duration property is set to a new value.

```js
effect.updateTiming({ iterations: 2 }); 
// Duration is undefined, and we should not clear the autoDuration flag.

effect.updateTiming({ duration: 200 });
// Duration is updated to a non-auto value, and we need to clear the flag.
```

We could clear the flag even if the duration is 'auto', as it will be set again when we renormalize time: 
```js
        if (typeof duration !== 'undefined') {
          details.autoDurationEffect = null
        }
```

I could open a separate PR to simplify that check?

On a side note, maybe we need better naming for the flag?
The flag is used to keep track of whether the user specified an auto-duration, as the 'auto' value will disappear when we normalize the time to INTERNAL_DURATION_MS in `getTiming()`.

> Furthermore, wouldn’t it be possible to do this check for `argumentsList[0]` earlier, and have it exit prematurely if `argumentsList.length` equals `0`?
> 
> I.e.
> 
> ```js
> apply: function(target, thisArg, argumentsList) {
>   if (!argumentsList || !argumentsList.length) return;
> 
>   if (details.timeline) {
>   …
> ```

Absolutely. I've updated the PR to do this instead.

bramus: I think this looks good for now. The mentioned refactor can be part of a separate PR.

PR #192, opened 2023-12-28
Schedule tasks for timeline updates

johannesodland:
Scheduling a task to update timelines, instead of waiting for the next animation frame. 

Some tests only wait for _one_ animation frame. If we wait for the next animation frame before ticking animations, it might result in assertions running right before the animations are ticked.

Tests that that use `waitForNextFrame()` from [web-animations/testcommon.js](https://github.com/web-platform-tests/wpt/blob/a95a859194890421952a3105e9b69789dbb4c58b/web-animations/testcommon.js#L177) such as [the last test in the scroll-animation-inactive-timeline.](https://github.com/web-platform-tests/wpt/blob/a95a859194890421952a3105e9b69789dbb4c58b/scroll-animations/scroll-timelines/scroll-animation-inactive-timeline.html#L137) are affected by this issue.

This test should now pass.

### DIFF
--- a/src/scroll-timeline-base.js
+++ b/src/scroll-timeline-base.js
@@ -233,9 +233,8 @@ function updateMeasurements(source) {
   if (details.updateScheduled)
     return;
 
-  requestAnimationFrame(() => {
-    // Defer ticking timeline to animation frame to prevent
-    // "ResizeObserver loop completed with undelivered notifications"
+  setTimeout(() => {
+    // Schedule a task to update timelines after all measurements are completed
     for (const ref of details.timelineRefs) {
       const timeline = ref.deref();
       if (timeline) {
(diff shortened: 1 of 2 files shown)


johannesodland:
Fixing this causes the first subtest in /scroll-animations/scroll-timelines/scroll-timeline-invalidation.html to pass every now and then, depending on timing.

This subtest sets the height of the first child element as part of the test. To pass this subtest property I think we need to observe the children of the source element.

@flackr 
You were originally opposed to observing the children, but without it some tests will be flaky. Are you ok with observing the children, or should we add other measures such as polyfill options to activate/deactivate observers first?

flackr:
> You were originally opposed to observing the children, but without it some tests will be flaky. Are you ok with observing the children, or should we add other measures such as polyfill options to activate/deactivate observers first?

I'm okay with observing all the direct children. It's a bit unfortunate that we have to do this but seems like an okay tradeoff.

flackr (approved):
LGTM

PR #193, opened 2023-12-28
Handle exceptions when auto-aligning start time

johannesodland:
It is currently possible to set values for the attachment range that are valid length-percentage values, but that are not supported in the polyfill yet. 

This will cause en error to be thrown when we try to auto-align the start time, preventing tickAnimation from completing as expected.

We should probably add a stricter validation when setting the range, but for now I think we should catch the exception and fall back to using the full animation range.

Merging this PR should prevent the timeouts that occurred in the WPT subtests in #153

### DIFF
--- a/src/proxy-animation.js
+++ b/src/proxy-animation.js
@@ -886,29 +886,50 @@ function autoAlignStartTime(details) {
     return;
   }
 
-  const previousRangeDuration = details.rangeDuration
+  const previousRangeDuration = details.rangeDuration;
+
+  let startOffset, endOffset;
+
 
   // 5. Let start offset be the resolved timeline time corresponding to the start of the animation attachment range.
   //    In the case of view timelines, it requires a calculation based on the proportion of the cover range.
-  const startOffset = CSS.percent(fractionalStartDelay(details) * 100)
+  try {
+    startOffset = CSS.percent(fractionalStartDelay(details) * 100);
+  } catch {
+    // TODO: Validate supported values for range start, to avoid exceptions when resolving the values.
+
+    // Range start is invalid, falling back to default value
+    startOffset = CSS.percent(0);
+    details.animationRange.start = 'normal';
+    console.warn("Exception when calculating start offset", e);
+  }
 
   // 6. Let end offset be the resolved timeline time corresponding to the end of the animation attachment range.
   //    In the case of view timelines, it requires a calculation based on the proportion of the cover range.
-  const endOffset = CSS.percent((1 - fractionalEndDelay(details)) * 100)
+  try {
+    endOffset = CSS.percent((1 - fractionalEndDelay(details)) * 100);
+  } catch {
+    // TODO: Validate supported values for range end, to avoid exceptions when resolving the values.
+
+    // Range start is invalid, falling back to default value
+    endOffset = CSS.percent(100);
+    details.animationRange.end = 'normal';
+    console.warn("Exception when calculating end offset", e);
+  }
 
   // Store the range duration, until we can find a spec aligned method to calculate iteration duration
   // TODO: Clarify how range duration should be resolved
-  details.rangeDuration = endOffset.value - startOffset.value
+  details.rangeDuration = endOffset.value - startOffset.value;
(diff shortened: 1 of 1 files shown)


src/proxy-animation.js
```diff
@@ -882,29 +882,45 @@ function autoAlignStartTime(details) {
...
+  let startOffset, endOffset;
+
 
   // 5. Let start offset be the resolved timeline time corresponding to the start of the animation attachment range.
   //    In the case of view timelines, it requires a calculation based on the proportion of the cover range.
-  const startOffset = CSS.percent(fractionalStartDelay(details) * 100)
+  try {
+    startOffset = CSS.percent(fractionalStartDelay(details) * 100);
+  } catch {
+    // Could not resolve the start offset
```
bramus, line 895: Do we also need a TODO here, similar to the `endOffset` parsing below?

bramus:
> It is currently possible to set values for the attachment range that are valid length-percentage values, but that are not supported in the polyfill yet.

What are some examples of these values?

> Merging this PR should prevent the timeouts that occurred in the WPT subtests in https://github.com/flackr/scroll-timeline/pull/153

Perfect!

johannesodland:
> What are some examples of these values?

Currently we don’t support relative units such as `em` or `vh` as we can not resolve them. They will parse fine as length-percentage values, but will not resolve to `px` values, yet.

src/proxy-animation.js
johannesodland: Probably.. 👍 
Updated with separate TODOs right now. 

The TODOs really belong in parseAnimationRange/parseTimelineRangeOffset but these are going to be refactored in #153 and I don't want to make merging any harder 😅

bramus:
> Currently we don’t support relative units such as `em` or `vh` as we can not resolve them. They will parse fine as length-percentage values, but will not resolve to `px` values, yet.

Maybe we should add a `console.warn()` to notify the developer about this. This would be helpful to help them debug, thereby reducing frustration along the way.

johannesodland:
> Maybe we should add a `console.warn()` to notify the developer about this. This would be helpful to help them debug, thereby reducing frustration along the way.

Maybe 🤔

I considered this, but this procedure can be run quite often and it would spam the logs. Maybe if we add a flag per timeline and only issue the warning once? 🤔

bramus:
> I considered this, but this procedure can be run quite often and it would spam the logs.

Hmm, good call. Spamming the console with message is infeasible indeed.

> Maybe if we add a flag per timeline and only issue the warning once?

Another way to bypass this could be to error out entirely for that specific Scroll-Timeline: instead of adding an animation with the default (wrong) timeline offset, remove the scroll-animation altogether.

johannesodland:
> Another way to bypass this could be to error out entirely for that specific Scroll-Timeline: instead of adding an animation with the default (wrong) timeline offset, remove the scroll-animation altogether.

Hmmm.. I think we can leave the timeline active, as the range is set and calculated for the animation and not the timeline. The timeline might be used for multiple animations, and we should leave the others playing.

I think we can either 
1. cancel the current animation after a `console.warn()`
2. reset the animation range start/end to the default "normal" after a `console.warn()`, leaving the animation playing. 

I'm not sure which would be the best, but I'm slightly leaning towards the latter.
It could look something like this:

```js

  try {
    startOffset = CSS.percent(fractionalStartDelay(details) * 100);
  } catch (e) {
    // TODO: Validate supported values for range start, to avoid exceptions when resolving the values.

    // Range start is invalid, falling back to default value
    startOffset = CSS.percent(0);
    details.animationRange.start = 'normal';
    console.warn("Exception when calculating start offset", e);
  }
```

It would be better to validate the range when it's set. It might be hard to detect whether a CSSNumericValue contains unsupported units, but we might be able to leverage  [CSSNumericValue toSum()](https://drafts.css-houdini.org/css-typed-om/#dom-cssnumericvalue-tosum) to check which units are used. This is something we should solve in a separate PR though.

johannesodland:
@bramus I implemented the change I suggested in the comment above.

PR #197, opened 2024-01-07
Handle explicitly undefined options argument in updateTimingHandler

johannesodland:
This is a follow up on #191.

Handle case where effect.updateTiming() is called with an explicit undefined options argument.

This happens in the [view-timeline-subject-size-changes test](https://github.com/web-platform-tests/wpt/blob/a95a859194890421952a3105e9b69789dbb4c58b/scroll-animations/view-timelines/view-timeline-subject-size-changes.html#L53).

After a discussion in #191 we skipped checking for an explicitly undefined options argument, and instead returned if the argumentList was undefined or empty. That does not cover the case where the method is called explicitly with undefined as an argument as below. 

```js
effect.updateTiming(undefined)
// argumentList == [undefined]
// arguementList.length == 1
```

### DIFF
--- a/src/proxy-animation.js
+++ b/src/proxy-animation.js
@@ -784,7 +784,7 @@ function createProxyEffect(details) {
         return;
 
       // Additional validation that is specific to scroll timelines.
-      if (details.timeline) {
+      if (details.timeline && argumentsList[0]) {
         const options = argumentsList[0];
         const duration = options.duration;
         if (duration === Infinity) {


johannesodland:
@bramus I realized our changes in #191 doesn't handle the case where `undefined` is passed explicitly as an argument to `updateTiming()`, as it is done in the WPT subtest. Do you mind reviewing this PR?

PR #198, opened 2024-01-07
Current time for inactive timeline

johannesodland:
Fix current time calculation for inactive timeline.

Depends on #193 and #192.


johannesodland:
The tests this PR was created to pass are currently passing. 
Closing.

PR #200, opened 2024-01-07
Avoid compressing constructor names when running tests

johannesodland:
Adding this PR for discussion. 

Microbundle compresses the build with terser by default. This results in CSSOM constructor names being replaced with `t` or similarly compressed names, which causes some subtests to fail. 

Subtests using the [css-typed-om test-helper.js](https://github.com/web-platform-tests/wpt/blob/master/css/css-typed-om/resources/testhelper.js#L52) fail, as the test-helper uses the constructor name to determine how to compare arguments.

I don't think microbundle allows for configuring terser directly. Configuration seems to be limited to enabling or disabling compression. The only available option I can see is building an uncompressed bundle for testing. Otherwise we would need to configure rollup, babel and terser ourself to ensure terser doesn't compress constructor names.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "build": "microbundle -f iife",
     "dev": "run-all \"serve\" \" microbundle watch -f iife \"",
     "deploy": "npm run build",
-    "test-setup": "node test/setup/checkout-wpt.mjs",
+    "test-setup": "node test/setup/checkout-wpt.mjs && npm run build -- --no-compress",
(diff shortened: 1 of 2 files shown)


bramus:
Feels a bit weird to run the tests with a different build, as there can be subtle differences that work / don’t work in the regular type of build. Ideally, the build used in tests is also the one that ships to users.

Or won’t this change affect regular users?

johannesodland:
Closing as this is addressed in https://github.com/flackr/scroll-timeline/issues/219, #224 and following PRs

PR #201, opened 2024-01-07
Handle case in simplify calculation where the root is a CSSMathProduct with one child

johannesodland:
Depends on #200 

Handle case in simplify calculation where the root has been simplified to a CSSMathProduct with one child.

If that child could be expressed in the canonical unit, it would have been simplified further, however if the child has a unit  such as percent that can't be resolved further, it will not be simplified.


johannesodland:
Closing, as this will be fixed by #223

PR #203, opened 2024-01-08
Fix unattached subject inset test

johannesodland:
This PR makes /scroll-animations/view-timelines/unattached-subject-inset.html pass.


src/scroll-timeline-base.js
```diff
@@ -590,7 +593,6 @@ export function getScrollParent(node) {
         return node;
     }
   }
-  return document.scrollingElement;
```
flackr, line 593: I don't think removing this default return is safe. I don't believe that we are guaranteed to reach the `document.scrollingElement` following the getContainingBlock chain. E.g. I think a fixed position element is not technically "contained" by the document.scrollingElement.

src/scroll-timeline-base.js
```diff
@@ -731,7 +733,13 @@ function parseInset(value) {
         throw TypeError(`Could not parse inset "${value}"`);
       }
     });
+  } else if (Array.isArray(value)) {
+    parts = value;
+  } else {
+    parts = [value];
   }
+
+
```
flackr, line 742: nit: The double newline feels a bit odd here.

flackr (changes requested):

src/scroll-timeline-base.js
johannesodland: Thanks. 🙏
Should be gone now.

src/scroll-timeline-base.js
johannesodland: Yeah..  
I realized it's better to just check if the node is connected or not before following the getContainingBlock chain.

flackr (approved):

flackr:
I'm pretty sure the failure is just a flaky test - I filed #231 to track landing a solution to avoid these failures.

PR #204, opened 2024-01-09
Polyfill CSS.supports(animation-timeline: …)

bramus:
There are may tests that rely on this check before even running the actual tests.

This PR polyfills `CSS.supports(animation-timeline: …)`, thereby allowing more tests to run.

The total number of tests now is 1080 vs 959 before. Checking https://wpt.fyi/results/scroll-animations?label=experimental&label=master&aligned there should be 1117 tests though.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -73,6 +73,14 @@ function initPolyfill() {
       "Error installing ScrollTimeline polyfill: could not attach WAAPI's getAnimations to document"
     );
   }
+
+  const oldSupports = CSS.supports;
+  CSS.supports = (ident) => {
+    if (ident.startsWith('animation-timeline:')) {
+      return true;
+    }
+    return oldSupports(ident);
+  };
 }
 
 initPolyfill();
--- a/src/index.js
+++ b/src/index.js
@@ -73,14 +73,6 @@ function initPolyfill() {
       "Error installing ScrollTimeline polyfill: could not attach WAAPI's getAnimations to document"
     );
   }
-
-  const origSupports = CSS.supports;
-  CSS.supports = (ident) => {
-    if (ident.includes('animation-timeline:')) {
-      ident = ident.replaceAll('animation-timeline', '--animation-timeline');
-    }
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -73,6 +73,14 @@ function initPolyfill() {
       "Error installing ScrollTimeline polyfill: could not attach WAAPI's getAnimations to document"
     );
   }
+
+  const origSupports = CSS.supports;
+  CSS.supports = (ident) => {
+    if (ident.includes('animation-timeline:')) {
```
flackr, line 79: There's probably no reason to check if the string includes animation-timeline, if we're only replacing animation-timeline. Also the includes check includes the trailing colon but the replace doesn't, is this a typo?

src/index.js
```diff
@@ -73,6 +73,14 @@ function initPolyfill() {
...
+      ident = ident.replaceAll('animation-timeline', '--animation-timeline');
```
flackr, line 80: We should ideally replace all polyfill supported properties, I think something like this:
```js
ident = ident.replaceAll(/(animation-timeline|scroll-timeline(-(name|axis))?|view-timeline(-(name|axis|inset))?|timeline-scope)\s*:/, '--supported-property:');
```

Also for encapsulation purposes we should only claim to support CSS properties if we've called initCSSPolyfill, i.e. this override of CSS.supports should be set up when initCSSPolyfill is called.

src/index.js
bramus: Removing check …

src/index.js
bramus: Yes, replacing all properties was the goal here, ergo this still being a draft PR. Will update to the regex you provided.

Going to move the logic into `initCSSPolyfill` to make sure it only plugs CSS.supports() when actually necessary.

bramus:
Addressed the concerns and updated the test results. Should be good to go, @flackr.

flackr (approved):

PR #206, opened 2024-01-21
Implement CSSNumericValue.parse()

johannesodland:
Add proper tokenization and parsing for CSSNumericValue.parse().

### DIFF
--- a/src/numeric-values.js
+++ b/src/numeric-values.js
@@ -1,9 +1,27 @@
+import {
+  CommaToken,
+  DelimToken,
+  DimensionToken,
+  FunctionToken, IdentToken,
+  LeftCurlyBracketToken,
+  LeftParenthesisToken,
+  LeftSquareBracketToken,
+  NumberToken,
+  PercentageToken, RightCurlyBracketToken,
+  RightParenthesisToken, RightSquareBracketToken,
+  Token,
+  tokenizeString,
+  WhitespaceToken
+} from './tokenizer';
+import {simplifyCalculation} from './simplify-calculation';
+
 /**
  * @typedef {{[string]: integer}} UnitMap
  * @typedef {[number, UnitMap]} SumValueItem
  * @typedef {SumValueItem[]} SumValue
  * @typedef {null} Failure
  * @typedef {{[string]: integer} & {percentHint: string | undefined}} Type
+ * @typedef {{type: 'ADDITION'}|{type: 'MULTIPLICATION'}|{type: 'NEGATE'}|{type: 'INVERT'}} ASTNode
  */
 
 const failure = null;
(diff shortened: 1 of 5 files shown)


test/expected.txt
```diff
@@ -905,6 +905,7 @@ PASS	/scroll-animations/view-timelines/block-view-timeline-current-time.tentativ
 FAIL	/scroll-animations/view-timelines/block-view-timeline-nested-subject.tentative.html	View timeline with subject that is not a direct descendant of the scroll container
 FAIL	/scroll-animations/view-timelines/change-animation-range-updates-play-state.html	Changing the animation range updates the play state
 FAIL	/scroll-animations/view-timelines/contain-alignment.html	Stability of animated elements aligned to the bounds of a contain region
+PASS	/scroll-animations/view-timelines/fieldset-source.html	Fieldset is a valid source for a view timeline
```
bramus, line 908: This test is flaky: it always passes for me locally but never when ran via the GitHub actions.

johannesodland: I just assumed it was a newly added test 🙈. 

It seems to have passed when ran via the GitHub actions now, or am I missing something?
https://github.com/flackr/scroll-timeline/actions/runs/7602264720/job/20702607225?pr=206#step:9:11

bramus: O wait, I’m wrong here. A run earlier today did include it. Ignore my original comment here.

bramus:
I think we might have come at a point where it might be a good idea to split off the CSSOM polyfill into its own repo with its own tests? I mean, feels like this could become its own library.

johannesodland:
> I think we might have come at a point where it might be a good idea to split off the CSSOM polyfill into its own repo with its own tests? I mean, feels like this could become its own library.

It still limited to implementing the parts needed for this library.. But I’ve been thinking we should start by moving the implementation into its own folder to make it more explicit that the code is meant for polyfilling parts of the typed cssom. 

As it is right now it could be tempting to use internal functions directly without going through the public api when implementing scroll driven animations. Then we would end up with too tight a coupling. 

I can create a refactoring PR once the open PRs are merged.

bramus:
> I can create a refactoring PR once the open PRs are merged.

Yeah, sounds like a good plan.

Which PRs specifically – other than this one – definitely need to be merged?

johannesodland:
> Which PRs specifically – other than this one – definitely need to be merged?

I was wrong. I looked through the other PRs, and it looks like none of them are touching the cssom implementation, so we should be fine after this PR is in. We could do it as part of this PR as well, if that's preferable.

bramus:
I think it’d be best to do this refactor as part of a separate PR.

src/simplify-calculation.js
```diff
@@ -84,7 +84,15 @@ export function simplifyCalculation(root, info) {
     if (root instanceof CSSUnitValue && root.unit === 'em' && info.fontSize) {
       root = new CSSUnitValue(root.value * info.fontSize.value, info.fontSize.unit);
     }
-    // 3. If root is a <calc-constant>, return its numeric value.
+    // 3. If root is a <calc-keyword> that can be resolved, return what it resolves to, simplified.
+    if (root instanceof CSSKeywordValue) {
+      //https://www.w3.org/TR/css-values-4/#calc-constants
+      if (root.value === 'e') {
+        return new CSSUnitValue(2.7182818284590452354, 'number');
```
bramus, line 91:
```suggestion
        return new CSSUnitValue(Math.E, 'number');
```

src/simplify-calculation.js
```diff
@@ -84,7 +84,15 @@ export function simplifyCalculation(root, info) {
...
+      } else if (root.value === 'pi') {
+        return new CSSUnitValue(3.1415926535897932, 'number');
```
bramus, line 93:
```suggestion
        return new CSSUnitValue(Math.PI, 'number');
```

src/numeric-values.js
```diff
@@ -405,4 +422,405 @@ export function multiplyTypes(type1, type2) {
...
+    }
+    const o2 = operatorStack.pop()
+    const second = tree.pop();
+    const first = tree.pop();
+    tree.push(toNAryAstNode(o2, first, second));
+  }
+  return tree[0];
+}
+
+function astToCSSNumericValue(node) {
```
bramus, line 687: I would add a comment here that this is step 4 from https://drafts.css-houdini.org/css-typed-om/#reify-a-math-expression. Took me some looking up to figure this out.

src/numeric-values.js
```diff
@@ -405,4 +422,405 @@ export function multiplyTypes(type1, type2) {
...
+}
+
+/**
+ * Reify a math expression
+ * https://drafts.css-houdini.org/css-typed-om/#reify-a-math-expression
+ * @param {CSSFunction} num
+ */
+function reifyMathExpression(num) {
+  // TODO: handle `clamp()` and possibly other math functions
+  // If num is a min() or max() expression:
```
bramus, line 736: In other files the step numbers of the algorithm was included in the source. I think it would be nice to add these here as well.

```suggestion
  // 1. If num is a min() or max() expression:
```

src/numeric-values.js
```diff
@@ -405,4 +422,405 @@ export function multiplyTypes(type1, type2) {
...
+  {
+    // Let values be the result of reifying the arguments to the expression, treating each argument as if it were the contents of a calc() expression.
+    const values = num.values
+      .filter(value => !(value instanceof WhitespaceToken || value instanceof CommaToken))
+      .map(value => reifyMathExpression(new CSSFunction('calc', value)));
+    // Return a new CSSMathMin or CSSMathMax object, respectively, with its values internal slot set to values.
+    return num.name === 'min' ? new CSSMathMin(...values) : new CSSMathMax(...values);
+  }
+
+  // Assert: Otherwise, num is a calc().
```
bramus, line 747:
```suggestion
  // 2. Assert: Otherwise, num is a calc().
```

src/numeric-values.js
```diff
@@ -405,4 +422,405 @@ export function multiplyTypes(type1, type2) {
...
+  if (num.name !== 'calc') {
+    return null;
+  }
+
+  // Turn num’s argument into an expression tree using standard PEMDAS precedence rules, with the following exceptions/clarification:
```
bramus, line 752:
```suggestion
  // 3. Turn num’s argument into an expression tree using standard PEMDAS precedence rules, with the following exceptions/clarification:
```

src/numeric-values.js
```diff
@@ -405,4 +422,405 @@ export function multiplyTypes(type1, type2) {
...
+  //
+  // Treat subtraction as instead being addition, with the RHS argument instead wrapped in a special "negate" node.
+  // Treat division as instead being multiplication, with the RHS argument instead wrapped in a special "invert" node.
+  // Addition and multiplication are N-ary; each node can have any number of arguments.
+  // If an expression has only a single value in it, and no operation, treat it as an addition node with the single argument.
+  const root = convertTokensToAST([...num.values]);
+  const numericValue = simplifyCalculation(astToCSSNumericValue(root));
```
bramus, line 759:
```suggestion
  
  // 4. Recursively transform the expression tree into objects
  const numericValue = simplifyCalculation(astToCSSNumericValue(root));
```

bramus (changes requested):
Some minor nits to address _(changes suggested)_.

src/numeric-values.js
johannesodland: Added a comment, and realized I had to reify `e` and `pi` here: 

https://github.com/flackr/scroll-timeline/commit/ee63815768397b3bdc38235909c1505290735fe8

johannesodland:
@bramus Thank you for the review 🙏
I think I have addressed all your remarks, and have rebased onto master.

bramus:
Thanks!


Pinging @flackr to take a look before merging. I think this one should go in before #221.

src/tokenizer.js
```diff
@@ -0,0 +1,792 @@
...
+  consume() {
+    if (this.returned.length) {
+      return this.returned.shift();
+    } else {
+      // console.log('Consuming', this.codePoints[this.index]);
+      return this.codePoints[this.index++];
+    }
+  }
+
+  reconsume(codePoint) {
```
flackr, line 141: It looks like reconsume is called with the codepoint that was just consumed. I assume it wouldn't make sense to reconsume something different. If so, can we just step index back instead of needing to maintain a separate returned list that needs to be considered by consume and peek?

src/tokenizer.js
```diff
@@ -0,0 +1,792 @@
...
+export class LeftCurlyBracketToken extends Token {}
+
+export class RightCurlyBracketToken extends Token {}
+
+class InputStream {
+  codePoints;
+  index = 0;
+  returned = [];
+  constructor(input) {
+    this.codePoints = Array.from(input).map(char => char.codePointAt(0));
```
flackr, line 129: I'm not sure how often this is called in the polyfill, but I suspect it would be more efficient to keep a reference to the original input and do the mapping to codepoints just in time as accessed by consume or peek.

flackr (commented):
Wow, this is a lot! I'm a bit concerned about the size increase. We'll have to think about what parts of this could be removed for a reduced build. Also, do we skip this parsing code if the browser supports CSSStyleValue.parse?

flackr:
> Wow, this is a lot! I'm a bit concerned about the size increase. We'll have to think about what parts of this could be removed for a reduced build. Also, do we skip this parsing code if the browser supports CSSStyleValue.parse?

To be clear, I'm ecstatic to see such a precise implementation following the spec steps! I just think the extra kilobytes will be too much for some sites, and we'll want to have a light version of the polyfill where you have to construct the CSSOM value instead of parsing it from a string.

src/tokenizer.js
johannesodland: It shouldn't be called outside the initial parsing, as we store the resulting CSSNumericValues, but that's no excuse 😅. I switched to iterating through the input string directly in https://github.com/flackr/scroll-timeline/pull/206/commits/6ddd897a7b1a9fc62883af224a569d1b1df7b502.

johannesodland:
> > Wow, this is a lot! I'm a bit concerned about the size increase. We'll have to think about what parts of this could be removed for a reduced build. Also, do we skip this parsing code if the browser supports CSSStyleValue.parse?
> 
> To be clear, I'm ecstatic to see such a precise implementation following the spec steps! I just think the extra kilobytes will be too much for some sites, and we'll want to have a light version of the polyfill where you have to construct the CSSOM value instead of parsing it from a string.

Yeah, I know it's a lot 😅, and by all means we don't have to use it. It's been an interesting exercise :)
The parsing code should only be executed in browsers without `CSSStyleValue.parse()` support (Looking at you Firefox), but it does add 2.1kB to the compressed build.

If we had a build system that could build two targets, we could build one with a full implementation of `CSSStyleValue.parse()` and a light build with with a simple or no implementation of `CSSStyleValue.parse()`. That simple implementation could be limited to supporting CSSUnitValues. 

The current solution though (that I'm at fault for), supports some math functions but can fail in more complex calculations. And there's no clear definition of what it will support or not. It also fails in browsers without regex lookahead support (Safari < 16.4).

src/tokenizer.js
johannesodland: Yeah.. I was worried that there were situations where a token was reconsumed out of order, but it doesn't seem like it. I've swapped to stepping the index back in https://github.com/flackr/scroll-timeline/pull/206/commits/6ddd897a7b1a9fc62883af224a569d1b1df7b502.

src/numeric-values.js
```diff
@@ -405,4 +423,410 @@ export function multiplyTypes(type1, type2) {
...
+  // If input is a list of CSS tokens, return input.
+  // If input is a list of CSS component values, return input.
+  if (Array.isArray(input)) {
+    return input;
+  }
+  // If input is a string, then filter code points from input, tokenize the result, and return the final result.
+  if (typeof input === 'string') {
+    return tokenizeString(input);
+  }
+  // Assert: Only the preceding types should be passed as input.
```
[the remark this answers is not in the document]
johannesodland: Yeah.. Fixed in https://github.com/flackr/scroll-timeline/pull/206/commits/c3b9c5b1ae33904d0f0f44560b3be888c9368675

src/numeric-values.js
```diff
@@ -405,4 +423,422 @@ export function multiplyTypes(type1, type2) {
...
+  throw new TypeError(`Invalid input type git${typeof input}`)
```
flackr, line 461: I think the text "git" prepended before the type may have been unintentional?

src/tokenizer.js
```diff
@@ -0,0 +1,801 @@
...
+      this.index -= String.fromCodePoint(codePoint).length
+    }
+  }
+
+  peek() {
+    const codePoints = []
+    let position = this.index
+    for (let i = 0; i < 3; i++) {
+      const nextCodePoint = this.input.codePointAt(position);
+      if (typeof nextCodePoint !== 'undefined') {
```
flackr, line 150: According to the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt), codePointAt only returns undefined if the index is out of range. I think it would be cleaner to change the for loop condition to `i < 3 && position < this.input.length`, then if the loop runs we always expect it to find a codepoint.

src/tokenizer.js
flackr: Thanks, this looks much simpler now!

src/tokenizer.js
flackr: Looks great!

src/numeric-values.js
johannesodland: Oh.. that's embarrassing 😅
I'm having an issue where the focus is delayed when switching between the IDE and the terminal 😞

Fixed in https://github.com/flackr/scroll-timeline/compare/8e7cf381430bd1d2c7511debec7d954ebaccaf66..4ed64d0898c93a488381f6ad35c42033b39698a8

src/tokenizer.js
johannesodland: Yeah.. That looks cleaner: 
Fixed here: https://github.com/flackr/scroll-timeline/compare/8e7cf381430bd1d2c7511debec7d954ebaccaf66..4ed64d0898c93a488381f6ad35c42033b39698a8

johannesodland: 🙏

flackr (approved):
Thanks for the changes, looks good to me!

PR #209, opened 2024-01-24
Add utility function to split string into component values

johannesodland:
When parsing component values passed as a string to `rangeStart`, `rangeEnd` and `inset` we have previously used str.split(), either with a single whitespace or with a regex with various issues:

- `'calc(2 * 10px) auto'.split(' ')` returns `['calc(2', '*', '10px)', 'auto']`, breaking apart functions
- `'calc(2 * 10px) auto'.split(/(?<!\([^\)]*)\s(?![^\(]*\))/)` returns `['calc(2 * 10px)', 'auto']` as expected, but causes an error in browsers without support for lookahead

This PR adds a utility function `splitIntoComponentValues(input)` that splits a string with multiple component values into an array of individual component value strings. These can then be processed either as keywords or be further parsed using `CSSNumericValue.parse()`.

```js
splitIntoComponentValues('calc(2 * 10px) calc(2 * 5px)');
// [('calc(2 * 10px)', 'calc(2 * 5px)']

splitIntoComponentValues('10px 10px').map(val => CSSNumbericValue.parse(val));
// [new CSSUnitValue(10, 'px'), new CSSUnitValue(10, 'px')]
```

### DIFF
--- a/src/proxy-animation.js
+++ b/src/proxy-animation.js
@@ -5,13 +5,12 @@ import {
   removeAnimation,
   fractionalOffset,
 } from "./scroll-timeline-base";
+import {splitIntoComponentValues} from './utils';
 
 const nativeDocumentGetAnimations = document.getAnimations;
 const nativeElementGetAnimations = window.Element.prototype.getAnimations;
 const nativeElementAnimate = window.Element.prototype.animate;
 const nativeAnimation = window.Animation;
-
-const rangeNameRegExp = new RegExp(`(${ANIMATION_RANGE_NAMES.join('|')})(?!-)`);
 
 class PromiseWrapper {
   constructor() {
@@ -926,7 +925,7 @@ function getNormalEndRange(timeline) {
   if (timeline instanceof ViewTimeline) {
     return { rangeName: 'cover', offset: CSS.percent(100) };
   }
-  
+
   if (timeline instanceof ScrollTimeline) {
     return CSS.percent(100);
   }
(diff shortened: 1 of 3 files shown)


johannesodland:
@bramus Rebased to account for changes from https://github.com/flackr/scroll-timeline/pull/153

src/proxy-animation.js
```diff
@@ -951,32 +950,32 @@ function parseAnimationRange(timeline, value) {
...
+
     parts.forEach(part => {
-      if (part.endsWith('%'))
-        offsets.push(parseFloat(part));
-      else
+      if (ANIMATION_RANGE_NAMES.includes(part)) {
         rangeNames.push(part);
+      } else {
+        offsets.push(CSSNumericValue.parse(part));
+      }
```
bramus, line 962: This piece of code can throw an exception.

For example, when an author sets the range to `'invalid 100%'`, the first component will be `'invalid'` and the second `'100%'`. When processing `'invalid'` the logic will fall through to the `CSSNumericValue.parse(part)` part, upon which it will throw an exception.

Further down below in the code there’s already little piece of similar code that catches the exception and throws a `TypeError`. It might be a good idea to do this here as well, so that it makes things easier for authors to debug.

```suggestion
      if (ANIMATION_RANGE_NAMES.includes(part)) {
        rangeNames.push(part);
      } else {
        try {
          offsets.push(CSSNumericValue.parse(part));
        } catch (e) {
          throw TypeError(`Could not parse range "${value}"`);
        }
      }
```

src/scroll-timeline-base.js
```diff
@@ -721,17 +721,14 @@ function parseInset(value) {
...
-        try {
-          return CSSNumericValue.parse(str);
-        } catch (e) {
-          throw TypeError('Invalid inset');
-        }
+      }
+      try {
+        return CSSNumericValue.parse(str);
+      } catch (e) {
+        throw TypeError('Invalid inset');
```
bramus, line 731: For more clarity we should include the inset to give author a clue.

```suggestion
        throw TypeError(`Could not parse inset "${value}"`);
```

src/utils.js
```diff
@@ -21,4 +21,54 @@ export function normalizeAxis(axis, computedStyle) {
...
+      } else if (char === ')') {
+        level -=1;
+      }
+      i++;
+    }
+    return input.slice(startIndex, i);
+  }
+
+  function consumeWhitespace() {
+    while (input.slice(i, i + 1) === ' ') {
```
bramus, line 61: Instead of checking for a space, we should check for any type of whitespace (`/\s/`).

src/utils.js
```diff
@@ -21,4 +21,54 @@ export function normalizeAxis(axis, computedStyle) {
...
+      i++;
+    }
+  }
+
+  while(i < input.length) {
+    if (input.slice(i, i + 1) === ' ') {
```
bramus, line 67: Instead of checking for a space, we should check for any type of whitespace (`/\s/`).

src/utils.js
```diff
@@ -21,4 +21,54 @@ export function normalizeAxis(axis, computedStyle) {
...
+export function splitIntoComponentValues(input) {
+  const res = [];
+  let i = 0;
+
+  function consumeComponentValue() {
+    let level = 0;
+    const startIndex = i;
+    while (i < input.length) {
+      const char = input.slice(i, i + 1);
+      if (level === 0 && char === ' '){
```
bramus, line 48: Instead of checking for a space, we should check for any type of whitespace (`/\s/`).

bramus (changes requested):
Overall structure looks good. A few remarks to address.

johannesodland:
> Overall structure looks good. A few remarks to address.

Thanks for the review @bramus 🙏 

I think I've addressed all your remarks: https://github.com/flackr/scroll-timeline/compare/1afa09f2c7a3c39efc4a348f0940fd6fb6b03085..8e1896c3c547119a6bafed25ffd764f0efd786b4

bramus (approved):

bramus:
Just came to the realization one can perfectly author something like `animation-range: calc(0% + 50px)calc(100% - 50px)` – with no spaces between the start and end – which isn’t handled by `splitIntoComponentValues()`.

Now, I don’t think anyone would write such a declaration and consider this be an extreme edge case, so I’m fine with merging this as is.

/ping @flackr for an opinion on this.

johannesodland:
> Just came to the realization one can perfectly author something like `animation-range: calc(0% + 50px)calc(100% - 50px)` – with no spaces between the start and end – which isn’t handled by `splitIntoComponentValues()`.

We could easily handle that in `splitIntoComponentValues()` when we consume the end parenthesis.

johannesodland:
Updated the PR to support `calc(...)calc(...)`:

https://github.com/flackr/scroll-timeline/compare/8e1896c3c547119a6bafed25ffd764f0efd786b4..14cde5bf6812b437282c49603ec79b10b8d702e4

PR #210, opened 2024-01-25
Account for insets when comparing size of subject to size of scrollport/view progress visibility range

johannesodland:
We should account for insets when evaluating `targetIsTallerThanContainer`. 
Currently the range is slightly off when the subject is close in size to the inset adjusted scrollport.

### DIFF
--- a/src/scroll-timeline-base.js
+++ b/src/scroll-timeline-base.js
@@ -669,6 +669,7 @@ export function calculateRange(phase, sourceMeasurements, subjectMeasurements, a
 
   let startOffset = undefined;
   let endOffset = undefined;
+  // Take inset into account when determining the scrollport size
   const adjustedScrollportSize = sizes.containerSize - inset.start - inset.end;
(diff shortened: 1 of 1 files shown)


src/scroll-timeline-base.js
```diff
@@ -669,7 +669,8 @@ export function calculateRange(phase, sourceMeasurements, subjectMeasurements, a
 
   let startOffset = undefined;
   let endOffset = undefined;
-  const targetIsTallerThanContainer = viewSize > sizes.containerSize ? true : false;
+  const adjustedScrollportSize = sizes.containerSize - inset.start - inset.end;
```
bramus, line 672: Might be good to add a comment here to why this is done:

```suggestion
  // Take inset into account when determining the scrollport size
  const adjustedScrollportSize = sizes.containerSize - inset.start - inset.end;
```

bramus (changes requested):

PR #211, opened 2024-01-25
Remove test summary from test output.

flackr:
Having the summary line leads to every PR which alters test expectations conflicting with every other PR which alters them. The summary can be computed by grepping the PASS / FAIL lines anyways so this should reduce the number of merge conflicts to only cases where the same sets of test expectations changed.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -54,7 +54,7 @@ jobs:
         run: "npm run test:wpt"
       - name: Expected results
         run: "npm run test:compare"
-      - name: Adding summary
+      - name: Test results summary
         run: echo "Passed $(grep -c '^PASS' ./test/report/summary.txt) of $(grep -c '^'  ./test/report/summary.txt) tests" >> $GITHUB_STEP_SUMMARY
       - name: Clean build files
         run: "rm -rf node_modules test/wpt"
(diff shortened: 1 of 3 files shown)


flackr:
@johannesodland I'm hoping this helps reduce the churn with multiple PRs

johannesodland:
Maybe we could add the summary line to the workflow summary?

```
      - name: Adding summary
        run: echo "Passed $(grep -c '^PASS' ./test/report/summary.txt) of $(grep -c '^'  ./test/report/summary.txt) tests" >> $GITHUB_STEP_SUMMARY
```

bramus:
Good call. It’s something that was causing some friction indeed.

I like Johannes’s suggestion to add an informative line to the build pipeline’s output.

bramus (changes requested):
Good call. It’s something that was causing some friction indeed.

I like Johannes’s suggestion to add an informative line to the build pipeline’s output.

flackr:
> Maybe we could add the summary line to the workflow summary?
> 
> ```
>       - name: Adding summary
>         run: echo "Passed $(grep -c '^PASS' ./test/report/summary.txt) of $(grep -c '^'  ./test/report/summary.txt) tests" >> $GITHUB_STEP_SUMMARY
> ```

Great suggestion, I've added this step as suggested.

.github/workflows/gh-pages.yml
```diff
@@ -54,6 +54,8 @@ jobs:
         run: "npm run test:wpt"
       - name: Expected results
         run: "npm run test:compare"
+      - name: Adding summary
```
bramus, line 57: Suggestion to reword the step name, as it doesn’t really “add” anything.

```suggestion
      - name: Test Results
```

bramus (approved):

PR #215, opened 2024-01-26
Speed up WPT checkout

bramus:
By shallow cloning and only cloning the master branch, the WPT setup can be sped up, thereby improving the overall speed of the build pipeline.

Without this change – and at the time of writing – 705377 files make up the entire WPT repo and its history. With it, only 89341 files need to be fetched.

See https://www.bram.us/2020/11/11/speed-up-build-times-with-this-little-git-trick/ for details about this trick.


flackr (approved):
LGTM

PR #217, opened 2024-01-28
Split JS and CSS loading of the polyfill

### LINKED ISSUE #160: Polyfill doesn't define `ScrollTimeline` at all.
Trying to use 
`import 'https://flackr.github.io/scroll-timeline/dist/scroll-timeline.js';`
Results in no ScrollTimeline defined. 
This is also a case on demo site: 
image

bramus:
Fixes #160.

Not sure if we do want to have this though, as it will leave Firefox Nightly somewhat useless state for CSS-based SDA’s.

With the changes in this PR, the polyfill won’t kick for the CSS part but will for the JS part in Firefox Nightly. However, because Firefox Nightly supports an older CSS syntax (e.g. using `vertical` instead of `y`, or requiring an time-based `animation-duration`) this will leave CSS Scroll-Driven Animations in a broken state.

One thing I do like about this PR is the code restructure that splits off the JS plugging into `scroll-timeline-js.js`.


bramus:
Converting to “draft” state to prevent accidental merges. Would love to get feedback on the situation described above though.

bramus:
RE Firefox Nightly's half-broken state: if the browser claims support then the polyfill should accept this as true. If their implementation is buggy, then they should fix that.

src/scroll-timeline-css.js
```diff
@@ -5,7 +5,7 @@ import { ScrollTimeline, ViewTimeline, getScrollParent, calculateRange,
```
flackr, line 5: Until we can get rid of the ScrollTimeline, ViewTimeline and ProxyAnimation imports I think the css polyfill will only work if the js polyfill was loaded. E.g. we use whether the returned animation is a ProxyAnimation to determine whether we need to set up a css scroll driven animation: https://github.com/flackr/scroll-timeline/blob/master/src/scroll-timeline-css.js#L157

We're similarly dependent on getAnimations being overridden to return the ProxyAnimation instances.

PR #221, opened 2024-02-01
Handle CSSNumericValue instances while parsing

### LINKED ISSUE #220: Uncaught TypeError: e.trim is not a function
I am getting the following error on Firefox (Version 122.0)

````
Uncaught TypeError: e.trim is not a function
    s proxy-cssom.js:178
    parse proxy-cssom.js:189
    _e proxy-animation.js:1044
    ze proxy-animation.js:1912
    <anonymous> (index):38
proxy-cssom.js:178:18
    s proxy-cssom.js:178
    parse proxy-cssom.js:189
    _e proxy-animation.js:1044
    ze proxy-animation.js:1912
    <anonymous> (index):38
````
Which happens in
````js
function parseCSSNumericValue(value) {
    value = value.trim();
    if (value.match(/^[a-z(]/i)) {
      return parseMathFunction(value);
    } else {
      return parseCSSUnitValue(value);
    }
 }
````

### PR DESCRIPTION
**Pull request by bramus**
Fixes #220.

### DIFF
--- a/src/proxy-cssom.js
+++ b/src/proxy-cssom.js
@@ -72,6 +72,8 @@ export function installCSSOM() {
 
   class CSSNumericValue {
     static parse(value) {
+      if (value instanceof CSSNumericValue) return value;
+
       return simplifyCalculation(parseCSSNumericValue(value), {});
     }
 
--- a/src/proxy-cssom.js
+++ b/src/proxy-cssom.js
@@ -247,7 +247,7 @@ export function installCSSOM() {
     });
   }
 
-  for (let [type, value] of Object.entries({cssOMTypes})) {
+  for (let [type, value] of Object.entries(cssOMTypes)) {
     if (type in window)
       continue;
     if (!Reflect.defineProperty(window, type, { value }))
--- a/src/proxy-cssom.js
+++ b/src/proxy-cssom.js
@@ -83,6 +83,8 @@ export function installCSSOM() {
   }
 
   const cssOMTypes = {
+    'CSSNumericValue': CSSNumericValue,
+    'CSSMathValue': CSSMathValue,
     'CSSUnitValue': class extends CSSNumericValue {
       constructor(value, unit) {
         super();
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by bramus** [comment]
Best to hold off merging this one until #206 makes it in, and then rebase these changes on top of that.

File: src/proxy-cssom.js [hunk 1]
```diff
@@ -350,10 +359,10 @@ export function installCSSOM() {
     });
   }
 
-  for (let type in cssOMTypes) {
+  for (let [type, value] of Object.entries({'CSSMathValue': CSSMathValue, 'CSSNumericValue': CSSNumericValue, ...cssOMTypes})) {
```
flackr, line 362: I think it would be cleaner to add these to cssOMTypes, e.g.
```js
const cssOMTypes = {
  'CSSMathValue': CSSMathValue,
  'CSSNumericValue': CSSNumericValue,
  /* existing code ... */
};

**Review by flackr — APPROVED** [review]

File: src/proxy-cssom.js (hunk 1 above)
Comment (bramus) [reply src/proxy-cssom.js:362]: I can't directly, because other classes extend those two classes. Could push them on the object though, right after declaring them but the net result is the same as the code I have right now.

Comment (bramus) [reply src/proxy-cssom.js:362]: No, it can't find the base class because it is not registered yet when declaring the new class that extends it.

Comment (flackr) [reply src/proxy-cssom.js]: Does the commit I pushed not work?

File: src/proxy-cssom.js [hunk 2]
```diff
@@ -197,6 +197,8 @@ export function installCSSOM() {
   }
 
   const cssOMTypes = {
+    'CSSNumericValue': CSSNumericValue,
+    'CSSMathValue': CSSMathValue,
```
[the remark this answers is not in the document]
bramus: Ah yes, I see like that. I was on a path where CSSNumericValue also got injected as an anonymous class. This works indeed.

src/proxy-cssom.js
bramus: Just saw that. That should work indeed. Build/Test is running.

src/proxy-cssom.js
```diff
@@ -359,7 +361,7 @@ export function installCSSOM() {
     });
   }
 
-  for (let [type, value] of Object.entries({'CSSMathValue': CSSMathValue, 'CSSNumericValue': CSSNumericValue, ...cssOMTypes})) {
+  for (let [type, value] of Object.entries({cssOMTypes})) {
```
flackr, line 364: Or rather, it would work if I'd removed the {} around cssOMTypes.

PR #223, opened 2024-02-03
Return "normal" for get rangeStart and rangeEnd

johannesodland:
We should return the string "normal" for rangeStart and rangeEnd (when specified as normal), and not a TimelineRangeOffset. 

This change makes `view-timeline-get-set-range.html` pass again, given that it is run on an uncompressed build (https://github.com/flackr/scroll-timeline/pull/200)  and that we have a proper implementation of CSSNumericValue.parse() (https://github.com/flackr/scroll-timeline/pull/206).

### DIFF
--- a/src/proxy-animation.js
+++ b/src/proxy-animation.js
@@ -6,6 +6,7 @@ import {
   fractionalOffset,
 } from "./scroll-timeline-base";
 import {splitIntoComponentValues} from './utils';
+import {simplifyCalculation} from './simplify-calculation';
 
 const nativeDocumentGetAnimations = document.getAnimations;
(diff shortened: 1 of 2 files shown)


bramus (approved):

bramus:
#206 and #225 are in, so the test results can get updated as part of this PR.

bramus (changes requested):
https://github.com/flackr/scroll-timeline/pull/206 and https://github.com/flackr/scroll-timeline/pull/225 are in, so the test results can get updated as part of this PR.

johannesodland:
@bramus Thanks for letting me know 🙏

I've rebased the branch, and the tests are now passing :)

flackr (approved):

Issue #219, opened 2024-01-31
Change build platforms for better configuration and multiple build targets


flackr:
This repo needs a build system that can handle the growing requirements. In particular, I think we need the following:
1. Multiple build targets. In particular, I would like to support:
   - a full build (i.e. roughly what's produced today),
   - a debug build (no mangling, easy to understand issues in dev tools debugging), 
   - one or more reduced builds, e.g.
      - a JS animation only build without the CSS support #173)
      - a build with minimal performance overhead, e.g. skips observers for size changes.

   I suspect the easy way we get different subsets of features is with different top level targets and modular includes. E.g. an include could install hooks for conformance code which would be no-ops when not installed.

2. Reserved name support. See #200 for a particular example where due to mangling we are failing tests.

3. Typescript? Not strictly required but I think having strict typing will make it easier to ensure correctness and make working on the polyfill easier.

@johannesodland @bramus WDYT? Does this seem good, anything else we need?

bramus:
4. (Stretch?) Auto publish tagged releases to NPM

PR #224, opened 2024-02-03
Use vite to build

### LINKED ISSUE #188: Run automated tests in strict mode.
In order to prevent issues like #183 we should make sure the automated tests are run with the polyfill loaded in strict mode.

flackr:
This is an experiment to see how vite works for #219. This PR right now only contains minimal changes to switch to vite. This fixes #188 as the build forces "use strict".


flackr:
This is looking pretty promising so far.

Going back to the list of goals, we should be able to:
1. Create multiple build targets. We can either use environment variables, or different vite config files to control this.
2. Reserved names can be specified by setting [build.terserOptions](https://vitejs.dev/config/build-options#build-terseroptions) to include a [reserved](https://terser.org/docs/options/#mangle-properties-options) list of names to not mangle.
3. Typescript support is as easy as invoking `tsc` before vite build.
4. I don't think auto-publishing to npm would be part of the build system. I think this would be better suited as a github action.
5. The github build action for vite seems to be 1s to microbundle's 4s. Vite also has a magically auto-updating dev server which reloads browser windows for extra speed.
6. As with #4, I think this is best met with github actions?

flackr:
@bramus @johannesodland I think we should go ahead with this, and then I can make followon PR's to meet the specific goals as outlined in https://github.com/flackr/scroll-timeline/pull/224#issuecomment-1925439103

bramus (approved):

PR #225, opened 2024-02-04
Don't mangle ScrollTimeline and ViewTimeline class names to satisfy wpt.

flackr:
@johannesodland @bramus This is an alternative to #200 now that we can use terser and configure exactly which classnames it shouldn't mangle. This addresses item 2 from #219.

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -10,7 +10,13 @@ export default defineConfig({
       name: 'ScrollTimeline',
       // the proper extensions will be added
       fileName: (format, entryAlias) => `scroll-timeline${format=='iife'?'':'-' + format}.js`,
-      formats: ['iife']
+      formats: ['iife'],
+    },
+    minify: 'terser',
+    terserOptions: {
+      mangle: {
+        keep_classnames: /(Scroll|View)Timeline/
+      }
     },
     rollupOptions: {
       output: {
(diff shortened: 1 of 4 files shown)


johannesodland:
Nice 😀🙏🏻

vite.config.js
```diff
@@ -10,7 +10,13 @@ export default defineConfig({
       name: 'ScrollTimeline',
       // the proper extensions will be added
       fileName: (format, entryAlias) => `scroll-timeline${format=='iife'?'':'-' + format}.js`,
-      formats: ['iife']
+      formats: ['iife'],
+    },
+    minify: 'terser',
+    terserOptions: {
+      mangle: {
+        keep_classnames: /(Scroll|View)Timeline/
```
flackr, line 18: Might even be good to just keep all classnames since as you pointed out in the other PR that some of the CSSOM names are important, it barely affects the size since there's not actually that many classes defined in the polyfill anyways.

flackr:
Here's a size comparison of the different options:

| keep_classnames | size |
| --- | --- |
| `false` | 51490 bytes |
| `/(Scroll\|View)Timeline/` | 51705 bytes |
| `true` | 51949 bytes |

Given the relatively minimal change to code size, it's probably fine to keep all classnames in the interest of correctness.

bramus (approved):

PR #226, opened 2024-02-04
Move typed om implementation into a separate folder

johannesodland:
Moving the internal implementation of typed om into its own folder.


bramus:
I’d also move `src/proxy-cssom.js` into that folder to contain all CSSOM related polyfilling into that folder. I think it would also make sense to out the several classes into their own separate files. This is something I also considered doing in #221.

johannesodland:
Yeah.. That makes sense. I didn't realize #221 was merged. 

I've moved `src/proxy-cssom.js` also now. 

Do you want me to split out some of the classes, or should we do that in a separate PR?

bramus:
It can go in this PR, as it's all about reorganization.

johannesodland:
Moved the token classes into a separate file now, when it comes to the CSSNumericValue classes, I think they should stay in proxy-cssom?

bramus:
Yes. That way, eventually, the whole cssom block can graduate to being its own package.

bramus:
Re-reading my comment I wasn't clear here. I interpreted "staying in CCSOM" as stay in the folder.

What I mean is: I'd split off these classes into their own files, one class per file.

johannesodland:
Got it :)

I split the classes into their own files and switched from using a WeakMap for the internals to using private class fields.

PR #228, opened 2024-02-04
Use simplicyCalculation to resolve px

johannesodland:
Using `simplifyCalculation()` in `resolvePx()` to resolve CSSMathValues and CSSUnitValues according to spec.
Swithcing to using `resolvePx()` elsewhere, in stead of calling `simplicyCalculation()` directly. 

This gives us one unified code path to resolving px values from CSSNumericValues. 

`resolvePx()` now takes an info/context object that `simplifyCalculation()` can use to resolve relative units. Ritght now it supports `percentageReference` and `fontSize`, but can be extended with values so that we can resolve all relative units, such as viewport units.

The context object is now called `info` but could benefit from a more descriptive name. Suggestions are welcome :)

### DIFF
--- a/src/scroll-timeline-base.js
+++ b/src/scroll-timeline-base.js
@@ -105,25 +105,17 @@ export function calculateMaxScrollOffset(source, axis) {
     return sourceMeasurements.scrollWidth - sourceMeasurements.clientWidth;
 }
 
-function resolvePx(cssValue, resolvedLength) {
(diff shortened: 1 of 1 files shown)


flackr (approved):
Looks good! I'm surprised this doesn't pass any new tests.

flackr:
Note that for each bit of information we pass in to info we technically should invalidate when that info changes.

PR #230, opened 2024-02-04
Fix scroll-animation.html start and current time subtests

johannesodland:
Schedule a micro task for `updatingInternal()` after adding an animation. If `updateInternal()` runs in the same task then `animation.currentTime` will be set by the time animation.play() returns.

Delaying this function revealed two other issues that needed to be fixed: 
- Infinite durations needed to be validated in getTimingHandler.
- holdTime was set to unresolved in tickAnimation, flipping the playState from 'finished' to 'running'.

### DIFF
--- a/src/scroll-timeline-base.js
+++ b/src/scroll-timeline-base.js
@@ -370,7 +370,9 @@ export function addAnimation(scrollTimeline, animation, tickAnimation) {
     animation: animation,
     tickAnimation: tickAnimation
   });
-  updateInternal(scrollTimeline);
+  queueMicrotask(() => {
+    updateInternal(scrollTimeline);
+  });
 }
 
 // TODO: this is a private function used for unit testing add function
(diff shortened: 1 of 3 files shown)


src/scroll-timeline-base.js
```diff
@@ -370,7 +370,9 @@ export function addAnimation(scrollTimeline, animation, tickAnimation) {
     animation: animation,
     tickAnimation: tickAnimation
   });
-  updateInternal(scrollTimeline);
+  queueMicrotask(() => {
```
flackr, line 373: I think if the timeline is active we expect the animation to tick immediately. It's only if this is a newly created timeline that hasn't ticked yet that we would expect it to wait until the next frame update to tick.

flackr: Though I think we also expect the animation to not start immediately due to the auto aligning start time behavior, so perhaps I'm worried about an edge case that isn't really web visible. We can revisit if this causes issues but it seems to be a net plus on tests.

flackr (approved):

PR #234, opened 2024-02-08
Fix regression due to handling of normal range

johannesodland:
Fix regression due to https://github.com/flackr/scroll-timeline/pull/223

We should return `normal` as the default  range start and end, but the parsing still expects a TimelineRangeOffset which can cause an issue.

Fixing the issue by only returning `normal` when no range was set by the user, keeping the TimelineRangeOffset from getNormalStart/EndRange otherwise.

There's already a todo to handle the full syntax later.

### DIFF
--- a/src/proxy-animation.js
+++ b/src/proxy-animation.js
@@ -959,13 +959,16 @@ function getNormalEndRange(timeline) {
 }
 
 function parseAnimationRange(timeline, value) {
-  const animationRange = {
-    start: 'normal',
-    end: 'normal',
-  };
-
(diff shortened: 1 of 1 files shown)


johannesodland:
@bramus Do you have time to review this?

bramus (approved):

bramus:
Merged it in.

Besides the current TODO’s for this code, passing in explicit `normal` values also doesn’t work properly. I have filed https://github.com/flackr/scroll-timeline/issues/236 for this.

PR #237, opened 2024-02-08
Realign start time after animation has been paused

johannesodland:
Another test where the spec and the wpt does not align. 
The wpt subtest tests that the start time is realigned on animation.play() after the animation has been paused. Chromes implementation does the same. 

I think it's best to follow the test and Chromes implementation here, and adjust later if necessary.


johannesodland:
What is the best approach here, when the spec, wpt tests and Chrome implementation differ?

Should we raise an issue in the csswg-issues repo? Or is it just that the tests and implementation is behind the spec?

PR #242, opened 2024-02-09
Parse and transpile external stylesheets - fixes #78.

### LINKED ISSUE #78: Support styles outside of style tag

flackr:
In this implementation, a stylesheet's href is replaced with an object URL to the updated stylesheet href.

### DIFF
--- a/src/scroll-timeline-css.js
+++ b/src/scroll-timeline-css.js
@@ -41,8 +41,26 @@ function initMutationObserver() {
     el.innerHTML = newSrc;
   }
 
-  function handleLinkedStylesheet(el) {
-    // TODO
+  function handleLinkedStylesheet(linkElement) {
+    // Filter only css links to external stylesheets.
+    if (linkElement.type != 'text/css' && linkElement.rel != 'stylesheet' || !linkElement.href) {
+      return;
+    }
+    const url = new URL(linkElement.href, document.baseURI);
+    if (url.origin != location.origin) {
+      // Most likely we won't be able to fetch resources from other origins.
+      return;
+    }
+    fetch(linkElement.getAttribute('href')).then(async (response) => {
+      const result = await response.text();
+      let newSrc = parser.transpileStyleSheet(result, true);
+      newSrc = parser.transpileStyleSheet(result, false);
+      if (newSrc != result) {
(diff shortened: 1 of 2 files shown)


src/scroll-timeline-css.js
```diff
@@ -41,8 +41,26 @@ function initMutationObserver() {
...
+    if (linkElement.type != 'text/css' && linkElement.rel != 'stylesheet' || !linkElement.href) {
+      return;
+    }
+    const url = new URL(linkElement.href, document.baseURI);
+    if (url.origin != location.origin) {
+      // Most likely we won't be able to fetch resources from other origins.
+      return;
+    }
+    fetch(linkElement.getAttribute('href')).then(async (response) => {
+      const result = await response.text();
```
calinoracation, line 55: Is it worthwhile to defer transpilation to a new task with a setTimeout to avoid blocking if input or other operations are in progress when the fetch completes?

flackr: I'm happy to put a `setTimeout(transpileFunction, 0)` in until we have a reason that it is necessary to transpile as soon as the request is done if you think this will be important.

flackr:
@calinoracation Does this look good as a first step and we can add the timeout and cross-origin whitelist as extra options? It might also be fine to make requests for all css files even if they might fail CORS.

calinoracation (approved):
Yeah this looks great. Totally in agreement on follow-up to either have an allow list or simply parse everything option and deal with the errors.

src/scroll-timeline-css.js
calinoracation: Yeah not critical here, I think I always try to have a performance outlook and know these requests happen pretty early in the user experience. It might definitely be a premature optimization though.

PR #244, opened 2024-02-11
Add support for timeline-scope property

johannesodland:
Adding initial support for the `timeline-scope` property, without support for the full cascade order. (Previous implementation did not support the cascade order either). 

This PR consists of three separate parts that were necessary to make the tests pass:
- Support for the `timeline-scope` property, and finding the referenced timeline in the timeline scope subtree
- Remove mapping for style elements that are removed from the dom
- Store 'auto' duration, so that it can be used when calculating effect timing

### DIFF
--- a/src/proxy-animation.js
+++ b/src/proxy-animation.js
@@ -803,7 +803,7 @@ function createProxyEffect(details) {
         }
 
         if (typeof duration !== 'undefined' && duration !== 'auto') {
-          details.autoDurationEffect = null
+          details.autoDurationEffect = false
         }
       }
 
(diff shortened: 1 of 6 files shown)


raphaelaleixo:
Any updates on this?

amomchilov:
This is great!

I built this myself, and vendored it on my site. Here's the steps I used:

```console
$ git clone https://github.com/flackr/scroll-timeline.git
$ cd scroll-timeline
$ git fetch origin pull/244/head:pr-244 pr-244
$ git checkout pr-244
$ npm install
$ npm run build
# use the ./dist/scroll-timeline.js
```

It would be really great if this got merged so this wasn't necessary, but it looks like this repo hasn't had any activity in 2 years :/

Issue #57, opened 2022-06-23
NPM support


souporserious:
First, thank you for such a great polyfill! The performance is much smoother than traditional scroll events. I was curious if this package will be available on NPM? I noticed `scroll-timeline-polyfill` doesn't exist in the registry, but it looks like everything is set up for this to work. I'm currently wanting to play with this polyfill in a React Codesandbox which uses create react app and doesn't seem to currently support URL imports. I was able to copy the dist file, but was curious if an NPM package will be supported to make things a little easier in these situations.

okyanusoz:
If you need help, I can publish the package.

Azarattum:
It would be nice if somebody were to publish and maintain this package.

curtgrimes:
Telling npm/yarn/pnpm to install from GitHub could also be an option if importing the src/index.js file works directly (I was able to install, but did not try going further because I realized this polyfill won't work for my use case):

```
pnpm i -D github:flackr/scroll-timeline
# or: 
pnpm i -D github:flackr/scroll-timeline#b3d05a5
```

dy:
@flackr any reason this is not published to npm?

calinoracation:
I started looking at this a bit and pulled in the package by SHA to start with. Since there's no bundled `dist/` folder in the repo now, that doesn't work incredibly well. I also needed to change the `microbundle` in the `build` script to not use `-f` to produce more formats, specifically my bundler was only happy using the `dist/scroll-timeline-polyfill.modern.js` output. 

If it were being published, it might make sense to move that to the `browser` field in `package.json` and make those changes to include `dist/` by default in the published tarball.

flackr:
I also have started looking into getting this set up. The rough plan is in PR https://github.com/flackr/scroll-timeline/pull/224#issuecomment-1925439103 with the build switch to vite. I think I should be able to set up a github action to publish releases to npm automatically following https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages

calinoracation:
Not sure if it's valuable but I've started testing publishing in a for-internal fork prior to the publishing y'all are going to do. I added some initial support to publish types with the package and a ES / CJS build.

souporserious:
Just noticed this library has been [published to NPM](https://www.npmjs.com/package/scroll-timeline-polyfill) 🎉 thank you @flackr!

ADTC:
_Comments moved to:_
* #296

ADTC:
[...]

PR #245, opened 2024-02-12
Add initial typescript support as well as bundling for ES & CJS

calinoracation:
This is a slight modification to support typescript type exporting as well as publishing ES & CJS versions along with the published build. It doesn't do auto publishing or tagging mentioned in https://github.com/flackr/scroll-timeline/issues/57, but might help with figuring a few things out in that process.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -21,20 +21,22 @@ import {
   elementGetAnimations,
   documentGetAnimations,
   ProxyAnimation
-} from "./proxy-animation.js";
+} from "./proxy-animation";
 
 import { initCSSPolyfill } from "./scroll-timeline-css"
 
-function initPolyfill() {
+export { ScrollTimeline, ViewTimeline }
+
+function initPolyfill(): void {
   // initCSSPolyfill returns true iff the host browser supports SDA
   if (initCSSPolyfill()) {
     return;
   }
 
-  if ([...document.styleSheets].filter((s) => s.href !== null).length) {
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -21,20 +21,22 @@ import {
...
+export { ScrollTimeline, ViewTimeline }
+
+function initPolyfill(): void {
   // initCSSPolyfill returns true iff the host browser supports SDA
   if (initCSSPolyfill()) {
     return;
   }
 
-  if ([...document.styleSheets].filter((s) => s.href !== null).length) {
+  if ([ ...document.styleSheets ].filter((s) => s.href !== null).length) {
```
calinoracation, line 36: I may of had the `pretty: true` flag turned on momentarily that led to this autoformatting, happy to remove it.

tsconfig.json
```diff
@@ -0,0 +1,29 @@
+{
+  "compilerOptions": {
+    "target": "ES2020",
```
calinoracation, line 3: All of these might not be necessary, I was working through which ones were needed or not but starting from a point where at least it was all working for us.

vite.config.js
```diff
@@ -25,4 +27,5 @@ export default defineConfig({
       },
     }
   },
+  plugins: [dts()],
```
calinoracation, line 30: This creates the `index.d.ts` file for the exported type information. Without this many folks importing through a bundler with typescript would require not-insignificant amounts of changes and no typing information. It's possible to do it without another package, but it helped simplify the output quite a bit.

Issue #158, opened 2023-09-18
Sites with even partial aphrodite usage break loading the polyfill


calinoracation:
We're in a case where we're migrating off of Aphrodite to more static CSS solutions, however we still have a significant amount of Aphrodite that hasn't been moved yet. Aphrodite gets really, really broken when you touch it's `<style>` tag. https://github.com/Khan/aphrodite/issues/376. I don't think there's much chance of it being fixed upstream, so am curious if it would make sense to build in a bail-out on that particular tag in this polyfill.

In the `scroll-timeline-css` file for `handleStyleTag`, I added this bail out to make it ignore them:
```tsx
if (el.innerHTML.trim().length === 0 || el.dataset.aphrodite === 'true') {
  return;
}
```

Not sure if this specific solution or something else should be considered, but it broke luckily just our staging environment before we found and fixed it. Even doing something ridiculous like `el.innerHTML = el.innerHTML` breaks Aphrodite.

flackr:
I agree we should have a way to skip the css parsing. The JS parts are also more efficient because they don't have scan anything to detect usage. I was thinking it might be nice to build multiple bundles so that you'd have a smaller script to include if you skip the CSS parsing.

flackr:
> ```tsx
> if (el.innerHTML.trim().length === 0 || el.dataset.aphrodite === 'true') {
>   return;
> }
> ```

Is this an attribute aphrodite sets? Would it make sense to land this as well?

calinoracation:
@flackr Ahh, yeah that sounds good, just hit this again adopting the new version, so would love to land this.

PR #249, opened 2024-02-16
Dont parse stylesheets controlled by aphrodite

calinoracation:
Fixes https://github.com/flackr/scroll-timeline/issues/158.

### DIFF
--- a/src/scroll-timeline-css.js
+++ b/src/scroll-timeline-css.js
@@ -32,7 +32,7 @@ function initMutationObserver() {
    */
   function handleStyleTag(el) {
     // Don’t touch empty style tags.
-    if (el.innerHTML.trim().length === 0) {
+    if (el.innerHTML.trim().length === 0 || el.dataset.aphrodite === 'true') {
       return;
     }
     // TODO: Do with one pass for better performance
--- a/src/scroll-timeline-css.js
+++ b/src/scroll-timeline-css.js
@@ -32,7 +32,7 @@ function initMutationObserver() {
    */
   function handleStyleTag(el) {
     // Don’t touch empty style tags.
-    if (el.innerHTML.trim().length === 0 || el.dataset.aphrodite === 'true') {
+    if (el.innerHTML.trim().length === 0 || 'aphrodite' in item.dataset) {
       return;
     }
     // TODO: Do with one pass for better performance
--- a/src/scroll-timeline-css.js
+++ b/src/scroll-timeline-css.js
@@ -31,7 +31,9 @@ function initMutationObserver() {
    * @param {HtmlStyleElement} el style tag to be parsed
    */
   function handleStyleTag(el) {
-    // Don’t touch empty style tags.
+    // Don’t touch empty style tags nor tags controlled by aphrodite.
+    // Details at https://github.com/Khan/aphrodite/blob/master/src/inject.js,
+    // but any modification to the style tag will break the entire page.
     if (el.innerHTML.trim().length === 0 || 'aphrodite' in item.dataset) {
       return;
     }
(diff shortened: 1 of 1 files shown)


src/scroll-timeline-css.js
```diff
@@ -32,7 +32,7 @@ function initMutationObserver() {
    */
   function handleStyleTag(el) {
     // Don’t touch empty style tags.
-    if (el.innerHTML.trim().length === 0) {
+    if (el.innerHTML.trim().length === 0 || el.dataset.aphrodite === 'true') {
```
flackr, line 35: This doesn't seem correct, in the [aphrodite code](https://github.com/Khan/aphrodite/blob/225f43c5802259a9e042b384a1f4f2e5b48094ea/src/inject.js#L38) they call
```js
styleTag.setAttribute("data-aphrodite", "");
```

This results in having the key set to the empty string so it does not evaluate equivalent to 'true'. I see the same issue on this [demo page](https://codesandbox.io/p/sandbox/css-in-js-demo-aphrodite-65bcs?file=%2Fsrc%2Findex.js) where if i get the style tag this still evaluates false.

calinoracation: Let me test our instance again, I feel like I had tested this before. Perhaps both or maybe I got it wrong, will update this Tuesday based on more tests in production.

calinoracation: Interesting, yeah I just re-tested and looked at a messaging URL we use and I get the following which is why it is working for us. Not sure if it's different build tools / settings or what, but maybe we can test for any existence of `data-aphrodite` and ignore it by default to avoid having to have a big list or special values.

```html
<style data-aphrodite="true" id="aphrodite" type="text/css">._1k81gub{position:re...</style>
```

calinoracation: view-source:https://www.airbnb.com/ for example

calinoracation: Updated to only check existence now

src/scroll-timeline-css.js
```diff
@@ -32,7 +32,7 @@ function initMutationObserver() {
    */
   function handleStyleTag(el) {
     // Don’t touch empty style tags.
```
flackr, line 34: Can we add to this comment explaining why we skip tags with aphrodite? Maybe link to https://github.com/Khan/aphrodite/blob/master/src/inject.js to show why we check for the aphrodite attribute.

src/scroll-timeline-css.js
flackr: Checking the existence sounds good to me. I don't know if the value "false" needs to be specifically handled - seems unlikely?

src/scroll-timeline-css.js
calinoracation: Yep, added!

src/scroll-timeline-css.js
```diff
@@ -31,8 +31,10 @@ function initMutationObserver() {
    * @param {HtmlStyleElement} el style tag to be parsed
    */
   function handleStyleTag(el) {
-    // Don’t touch empty style tags.
-    if (el.innerHTML.trim().length === 0) {
+    // Don’t touch empty style tags nor tags controlled by aphrodite.
+    // Details at https://github.com/Khan/aphrodite/blob/master/src/inject.js,
+    // but any modification to the style tag will break the entire page.
+    if (el.innerHTML.trim().length === 0 || 'aphrodite' in item.dataset) {
```
flackr, line 37: I think s/item/el ? I don't see item defined

src/scroll-timeline-css.js
calinoracation: Seems unlikely to me as well, probably not worth the added complexity.

src/scroll-timeline-css.js
calinoracation: omg Tuesday is getting me, and my over-reliance on typescript. Thank you, updated!

flackr (approved):

PR #250, opened 2024-02-17
Pass idlharness.window.html

johannesodland:


src/scroll-timeline-base.js
```diff
@@ -405,24 +405,10 @@ export class ScrollTimeline {
...
-    updateSource(this, element);
-    updateInternal(this);
-  }
-
   get source() {
     return scrollTimelineOptions.get(this).source;
   }
 
-  set axis(axis) {
-    if (!isValidAxis(axis)) {
```
bramus, line 418: In the updated code the value for axis is no longer validated. Can we re-add this somehow?

I.e. we would need to make sure that the polyfill throws just like the native `ScrollTimeline`:

```js
const st = new ScrollTimeline({ axis: 'invalid' });
VM232:1 Uncaught TypeError: Failed to construct 'ScrollTimeline': Failed to read the 'axis' property from 'ScrollTimelineOptions': The provided value 'invalid' is not a valid enum value of type ScrollAxis.
    at <anonymous>:1:12
```

johannesodland: I think axis is still validated when passed as an option to ScrollTimeline?

https://github.com/flackr/scroll-timeline/blob/bad2e068f7eb53538cddfd8d9d0bd45539aa77a8/src/scroll-timeline-base.js#L395-L403

The axis property/attribute is specced as readonly:
https://drafts.csswg.org/scroll-animations-1/#scrolltimeline

bramus: Ah, I see. You can’t set the axis so removing this chunk seems fine indeed.

PR #257, opened 2024-04-04
Force parsing the linked stylesheet before animation starts

DeltaEvo:
There is a chance for the animation to start before the polyfill parses the linked stylesheet because we download and parse it at the same time than the browser.

When this happens, the ScrollTimeline is not created because the associated options are not available yet (
getAnimationTimelineOptions return null).

To prevent this from happening we remove the href of the linked stylesheet to ensure that the polyfill will process the stylesheet before the browser.

### DIFF
--- a/src/scroll-timeline-css.js
+++ b/src/scroll-timeline-css.js
@@ -53,7 +53,14 @@ function initMutationObserver() {
       // Most likely we won't be able to fetch resources from other origins.
       return;
     }
-    fetch(linkElement.getAttribute('href')).then(async (response) => {
+
+    // Stop loading the stylesheet to ensure that we finished parsing it before it's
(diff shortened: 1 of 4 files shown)


src/scroll-timeline-css.js
```diff
@@ -53,14 +53,23 @@ function initMutationObserver() {
       // Most likely we won't be able to fetch resources from other origins.
       return;
     }
-    fetch(linkElement.getAttribute('href')).then(async (response) => {
+
+    // Stop loading the stylesheet to ensure that we finished parsing it before it's
+    // loaded by the browser (and that animations are only started after
```
jhoopmann, line 58: at this point it's mostly already parsed by the browser and interpreted, only the layouting isn't fully done yet.

PR #268, opened 2024-05-24
Update README.md with helpful note

mendelgordon:
I couldn't figure out why the polyfill wasn't working for me until I checked the actual code of the polyfill and saw that it doesn't work with css from other origins, so I think adding a note explaining this in the readme is pretty important.

Here's [the part of the polyfill](https://github.com/flackr/scroll-timeline/blob/702b716afa35bd2b948a4b208ed552db2472de8e/src/scroll-timeline-css.js#L51) I'm referring to:
```javascript
const url = new URL(linkElement.href, document.baseURI);
if (url.origin !== location.origin) {
  // Most likely we won't be able to fetch resources from other origins.
  return;
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,10 @@ Also works with CSS Animations that use a `view-timeline` or `scroll-timeline`
 }
 ```
 
+Please ensure your CSS is hosted on the same domain as your website or included directly on the page within a <style> tag.
+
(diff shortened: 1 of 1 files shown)


flackr:
Thanks for the added note! I know it seems silly for a note in the README, but would you be willing to sign the CLA? https://github.com/flackr/scroll-timeline/blob/master/CONTRIBUTING.md

mendelgordon:
Np- done

PR #269, opened 2024-05-30
Create "lite" version (js only, without CSS)

egirard:
Also changes test for CSS support to look for view-timeline


vite.config.lite.js
```diff
@@ -0,0 +1,29 @@
...
+    rollupOptions: {
+      output: {
+        // Provide global variables to use in the UMD build
+        // for externalized deps
+        globals: {
+        },
+      },
+    }
+  },
+})
```
flackr, line 29: This is duplicating a lot of config that likely should be kept in sync. Can you move all of this to a file like `vite.common.js`, e.g.
```js
export function buildConfig(source, filename) {
    return {
        build: {
            sourcemap: true,
            emptyOutDir: false,
            lib: {
            // Could also be a dictionary or array of multiple entry points
            entry: source,
            name: 'ScrollTimeline',
            // the proper extensions will be added
            fileName: (format, entryAlias) => `${filename}${format=='iife'?'':'-' + format}.js`,
            formats: ['iife'],
            },
            minify: 'terser',
            terserOptions: {
            keep_classnames: /^((View|Scroll)Timeline)|CSS.*$/
            },
            rollupOptions: {
            output: {
                // Provide global variables to use in the UMD build
                // for externalized deps
                globals: {
                },
            },
            }
        }
    };
}
```

vite.config.js:
```js
import { resolve } from 'path'
import { defineConfig } from 'vite'
import { buildConfig } from './vite.common'

export default defineConfig(buildConfig(resolve(__dirname, 'src/index.js'), 'scroll-timeline'));
```

vite.config.lite.js:
```js
import { resolve } from 'path'
import { defineConfig } from 'vite'
import { buildConfig } from './vite.common'

export default defineConfig(buildConfig(resolve(__dirname, 'src/index-lite.js'), 'scroll-timeline-lite'));
```

This also eliminates the need to have a separate `dist` dir by configuring vite to not empty the dir on building.

src/index.js
```diff
@@ -25,47 +25,17 @@ import {
 
 import { initCSSPolyfill } from "./scroll-timeline-css"
 
-function initPolyfill() {
+import { initPolyfill } from "./init-polyfill.js"
+
+function initPolyfillIncludingCSS() {
   // initCSSPolyfill returns true iff the host browser supports SDA
+  console.debug("Polyfill initializing.");
   if (initCSSPolyfill()) {
```
flackr, line 33: I'm a bit unsure about this. If the browser supports the CSS API's but not the JS ones maybe we should still polyfill the JS apis.

flackr (changes requested):
I think it's just churn in the wpt repo, but it looks like there's two fewer tests run than expected: https://github.com/flackr/scroll-timeline/actions/runs/9304178220/job/25608156082?pr=269

Can you update expected.txt?

src/index.js
bramus: Best to check with `typeof ScrollTimeline !== "undefined"` (and `ViewTimeline`) here indeed.

src/scroll-timeline-css.js
```diff
@@ -154,7 +154,7 @@ function updateKeyframesIfNecessary(anim, options) {
 
 export function initCSSPolyfill() {
   // Don't load if browser claims support
-  if (CSS.supports("animation-timeline: --works")) {
+  if (CSS.supports("view-timeline")) {
```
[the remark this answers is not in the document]
egirard: I'm going to pull this mod out of the current patch and land it in a separate CL - I think there are several places where we have non-working feature detection.

src/init-polyfill.js
```diff
@@ -0,0 +1,69 @@
...
+  elementGetAnimations,
+  documentGetAnimations,
+  ProxyAnimation
+} from "./proxy-animation.js";
+
+import { initCSSPolyfill } from "./scroll-timeline-css"
+
+export function initPolyfill() {
+  // Don't load if browser claims support
+  if (CSS.supports("view-timeline")) {
```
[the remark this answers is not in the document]
egirard: Done.

src/index-lite.js
```diff
@@ -0,0 +1,29 @@
...
+
+import {
+  ScrollTimeline,
+  ViewTimeline,
+} from "./scroll-timeline-base";
+import {
+  animate,
+  elementGetAnimations,
+  documentGetAnimations,
+  ProxyAnimation
```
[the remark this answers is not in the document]
egirard: Done

vite.config.lite.js
egirard: Done.

egirard:
Updated code to address feedback - thanks.
Note that the css feature detection is now removed and I'll reland in a new patch.

.gitignore
```diff
@@ -21,6 +21,7 @@ node_modules
 # demos deployment directory
 build
 dist
+dist-lite
```
flackr, line 24: This can be removed - there's no dist-lite now.

src/index-lite.js
```diff
@@ -0,0 +1,17 @@
+// Copyright 2019 Google LLC
```
flackr, line 1: I think the convention is to put the year in which this file was added.

demo/view-timeline/index.html
```diff
@@ -169,7 +169,7 @@
   <input type="checkbox" name="timeline-insets" id="timeline-insets">
   <label for="timeline-insets">Use inset 100px 200px</label>
 </body>
-<script src="../../dist/scroll-timeline.js"></script>
+<script src="../../dist-lite/scroll-timeline-lite.js"></script>
```
flackr, line 172: this is just dist/scroll-timeline-lite.js now.

flackr (approved):
Looks good with a few nits.

demo/view-timeline/index.html
egirard: Thanks - fixed.

PR #274, opened 2024-07-10
Document lite version and npm

egirard:
Update readme to cover npm and lite version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,12 +4,27 @@ A polyfill of ScrollTimeline and ViewTimeline as defined by the [spec](https://d
 
 View a [cool demo showing its usage](https://flackr.github.io/scroll-timeline/demo/parallax/)!
 
+# News
+
+Recent updates:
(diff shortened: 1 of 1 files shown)


ivodolenc:
Hi @egirard , I just bumped into this, since there are merged PR's that are supposed to separate the JS related code and the rest.

But looking at the import files, that doesn't make much sense since those calls are basically the same?

`initPolyfill()` are called in `index.js` and also in the `index-lite.js` (as well as `proxy-animation` and `scroll-timeline-css`).

Maybe I missed something here, but this code does not separate modules as one would expect, or at least I would expect this to be just an import `scroll` and `view` JS polyfill without `proxy-animation` and other similar code.

For proper code separation, I believe this should be completely rewritten, but some features that use `CSS OM`, such as animation `ranges` and others, cannot be implemented separately due to `Timelines` limits and design choices.

The correct way would be to just mention these things in the readme and inform users, no need for additional files.

Issue #261, opened 2024-04-14
CSS src url changes relative paths to root


mevanloon:
I have some font faces and image that are referenced relatively to the stylesheet, but those are all converted to "root" paths. So `assets/fonts/coolfont.woff2` would become `/assets/fonts/coolfont.woff2`, which makes it not load.

xenoborg:
My fonts work fine

@font-face {
    font-family: alumni-sans;
    src: url('fonts/Alumni_Sans/AlumniSans-VariableFont_wght.ttf');
}

but using the polyfill breaks any background image url or mask image url only in Firefox though, requiring a full web url to fix eg www.mywebsite.co.uk/folder/folder/i/test.png

.content__image {
width: 300px;
height: 300px;
clip-path: polygon(10% 0, 100% 0, 100% 100%, 0 100%);
background-position: center;
background-image: url('i/test.png');
background-repeat: no-repeat;
background-color: black;
background-size: contain;
}

weisJ:
I noticed this too. Background images are broken in Firefox (Linux only).

weisJ:
This is due to the polyfill for non-inline styles putting the changed stylesheet into blob storage, which seems to have issues in firefox with relative urls. Seems like a non-css version would be very beneficial here. Or at least don't change stylesheets which don't need polyfilling.

mevanloon:
Indeed, it would be cool if there was a workaround where we could select what stylesheets to parse, or which ones to ignore. I experience this in webkit btw.

robzor:
I am also encountering this with my fonts, the polyfill works fine on Safari but not Firefox. 

This the path which has to be in this format for our Vite config to work:

```
src: url('/fonts/greed-condensed/greed-condensed-heavy.woff2') format('woff2');
```

Does anyone have a fix or a workaround for this? @flackr 

Thanks,

Robin

robzor:
@bramus do you have any thoughts on this? You seem to be active in the issues! :)

benefiction:
> I am also encountering this with my fonts, the polyfill works fine on Safari but not Firefox.
> 
> This the path which has to be in this format for our Vite config to work:
> 
> ```
> src: url('/fonts/greed-condensed/greed-condensed-heavy.woff2') format('woff2');
> ```
> 
> Does anyone have a fix or a workaround for this? @flackr
> 
> Thanks,
> 
> Robin

Hi Robin,

[...]

PR #279, opened 2024-08-08
Resolve relative urls for blob stylesheets

manuelmeister:
Solve #261

Added an additional demo to see the relative urls in action


src/scroll-timeline-css-parser.js
```diff
@@ -84,9 +84,17 @@ export class StyleParser {
...
-    return p.sheetSrc;
+    // Otherwise, we have to find `url()` functions and resolve
+    // relative URLs to absolute URLs.
+    if (!srcUrl) {
+      return p.sheetSrc;
+    }
+
+    const srcUrlDir = srcUrl.lastIndexOf('/') > location.origin.length
+        ? srcUrl.substring(0, srcUrl.lastIndexOf('/'))
+        : location.origin;
```
flackr, line 95: Why use location.origin instead of using the same srcUrl.substring expression here?

manuelmeister: Because I thought it could possibly resolve to just `https://`

manuelmeister: If we load the css from `https://subdomain.example.com` exactly (and the stylesheet is the response)

src/scroll-timeline-css-parser.js
```diff
@@ -84,9 +84,17 @@ export class StyleParser {
...
+
+    return p.sheetSrc.replace(/url\((["'])?(?:\.?\/|(?!https?:\/\/|(?:data|blob):))/gm, `url($1${srcUrlDir}/`);
```
flackr, line 97: Correct me if I'm wrong, but I think there are a couple issues with this replacement:
1. The optional dot `\.?` seems like it will treat `url('./foo.png')` and `url('/foo.png')` the same, but shouldn't the latter be relative to the root of the associated url?
2. It seems like it replaces quoted data URI's, e.g. try it with `url('data:text/css,bar')`, it inserts the `srcUrlDir` before the quote.

demo/view-timeline-external-css-with-url/styles.css
```diff
@@ -0,0 +1,48 @@
...
+  animation-timeline: foo;
+  animation-range: entry 0% 100%;
+  animation: linear anim;
+}
+*/
+
+@keyframes anim {
+  from { top: 0px; left: 0px; }
+  to { top: 100px; left: 200px; }
+}
```
flackr, line 48: Odd, when I load this page in Firefox or Safari I run into an error parsing the CSS so I'm unable to see how this works:
Unhandled Promise Rejection: Error: (<anonymous file>): Advanced beyond the end

src/scroll-timeline-css-parser.js
flackr: Ah that makes sense, but I know there have been requests to support loading from other origins when CORS is enabled - see #248 - so it'd be great to still use the srcUrl.

flackr (commented):
Thanks for tackling this! This looks really close though I seem to be unable to test it - see my comment. I'm curious if you're not getting the same issue.

flackr:
Also note we require a signed CLA to accept contributions, there's a link at https://github.com/flackr/scroll-timeline/blob/master/CONTRIBUTING.md#contributor-license-agreement with instructions on how to sign it.

demo/view-timeline-external-css-with-url/styles.css
manuelmeister: Sorry, I can't reproduce this (anymore?)

src/scroll-timeline-css-parser.js
manuelmeister: Fixed in 24ba542

src/scroll-timeline-css-parser.js
manuelmeister: Makes sense. I now use the origin of srcUrl.

manuelmeister:
@flackr What is left to do in this PR?

manuelmeister:
@flackr can you have a look at my fixes?

manuelmeister:
@flackr can I do something more to get this PR merged?

demo/view-timeline-external-css-with-url/styles.css
flackr: Looks like this has something to do with `vite` rewriting css. The response for the request looks nothing like the original css. This is obviously not realated to this patch, sorry for the noise.

PR #288, opened 2024-10-08
Handle double precision in playState

johannesodland:
Double precision errors can prevent `playState` from becoming "finished",
which in turn prevents the finished promise from resolving.

This is caused by precision loss due to floating point computations. For instance, converting an internal `holdTime` of 100000 to CSSNumberish and back can cause `currentTime` to become `99999.99999999999`. 

When the imprecise `currentTime` is compared against the `effectEnd()` with a regular `>=` operator the comparison will fail. This PR introduces an epsilon check to address this and ensure that the `playState` can become "finished".

### DIFF
--- a/src/proxy-animation.js
+++ b/src/proxy-animation.js
@@ -1078,6 +1078,20 @@ function parseTimelineRangePart(timeline, value, position) {
   unsupportedTimeline(timeline);
 }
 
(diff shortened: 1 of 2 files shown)


porfirioribeiro (approved):
This may be the reason why sometimes my timelines never finish!

Issue #298, opened 2025-03-21
I manually fixed 2 errors, here what i done, and what i found to help you patch the errors correctly.


RsMan-Dev:
I am working in solidjs, and many elements are manipulated, so for the following errors, i cannot get easily the original cause, if you cannot reproduce my bugs, i can help you to search.

Sometimes, scroll timeline tries to get progress when clientHeight and scrollHeight are same value, so when the progress is calculated, we get a 0/0, that give us Infinity, Infinity - Infinity gives NaN, and NaN goes up to your `setNativeCurrentTime` function, where ` details.animation.currentTime` gets `NaN - 0.001`, an error is thrown.

I patched it in your compiled code temporally using:
```js
    // patches : <v>.animation.currentTime = <a> + <t>
    // to : <v>.animation.currentTime = (<a> + <t>) || 0
    // will avoid NaN to pass, and will set 0 if the result is NaN
    scrollTimelinePolyfill = scrollTimelinePolyfill.replace(
      /(?<vName>[^.]+)\.animation\.currentTime\s*=\s*(?<aName>[^\s+]+)\s*\+\s*(?<tName>[^\s}]+)/,
      (...match) => {
        const {vName, aName, tName} = match.at(-1)
        return `${vName}.animation.currentTime = (${aName} + ${tName}) || 0`
      }
    )
```

Sometimes `measureSource` can get an undefined value as source, when `updateMeasurements` is called from ViewTimeline's constructor's ResizeObservers, `details.source` is undefined, probably due to DOM change from my app that made this error, but don't know what edit made it as the framework i'm using is handling the DOM updates itself.

I patched it in your compiled code temporally using:
```js
    // patches a : new ResizeObserver(() => {<f>(<v>.source)}).[...]
    // to : new ResizeObserver(() => {<v>.source && <f>(<v>.source)}).[...]
    // will avoid error when <v>.source is undefined, we are ignoring the error
    scrollTimelinePolyfill = scrollTimelinePolyfill.replace(
      /new\s*ResizeObserver\(\s*\(?\s*\(\s*\)\s*=>\s*{\s*(?<fName>[^(]*)\(\s*(?<vName>[^(]*)\.source\s*\)\s*}\s*\)?\s*\)\s*\./,
      (...match) => {
        const {fName, vName} = match.at(-1)
        return `new ResizeObserver((() => {${vName}.source && ${fName}(${vName}.source)})).`
      }
    )
```

PR #302, opened 2025-08-29
Fix: De-activate animation when scroll container has no overflow

dnlzro:
## Problem

This polyfill applies scroll-driven animations even when the scroll container has no overflow (i.e., when `clientHeight === scrollHeight`). According to the specification ([source](https://drafts.csswg.org/scroll-animations-1/#scroll-timeline-progress)):

> Progress (the current time) for a scroll progress timeline is calculated as: *scroll offset* ÷ (*scrollable overflow size* − *scroll container size*)
> 
> If the 0% position and 100% position coincide (i.e. the denominator in the current time formula is zero), the timeline is inactive.

When there is no overflow, then *scrollable overflow size* − *scroll container size* = 0. Currently, the timeline is set to 100% or `NaN` in these cases. Instead, it should be set to inactive.

## Changes

1. **`ScrollTimeline.currentTime`**: Return `unresolved` when `maxScrollPos <= 0`.
2. **`ViewTimeline.currentTime`**: Return `unresolved` when `offsets.end === offsets.start`.

**Note:** When the `currentTime` method returns `unresolved` (i.e., `null`), the `tickAnimation` function will cancel the animation. This ensures no effects are applied when the timeline is inactive.

**Note:** I think these changes also indirectly address the first issue raised in #298.

### DIFF
--- a/src/scroll-timeline-base.js
+++ b/src/scroll-timeline-base.js
@@ -473,7 +473,7 @@ export class ScrollTimeline {
     const maxScrollPos = calculateMaxScrollOffset(container, axis);
 
     return maxScrollPos > 0 ? CSS.percent(100 * scrollPos / maxScrollPos)
-                            : CSS.percent(100);
+                            : unresolved;
   }
 
   get __polyfill() {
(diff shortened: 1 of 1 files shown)


boneyfantaseas:
Instantly no annoying "Animation.currentTime" errors anymore, lovely. I use it for a scrollbar indicator approach and I learned that it works this solution works with indicators as pseudo elements but not as real DOM nodes.

